PR #7, opened 2022-10-18
Solution implemented

OlehMy:


oksana-logos-frontend (approved):
Good job🔥

src/createServer.js
```diff
@@ -1,3 +1,86 @@
...
+  const errorTypes = {
+    textIsMissing: 'text is missing',
+    toCaseIsMissing: 'to case is missing',
+    notSupportedCase: 'not supported case',
+  };
+  const errorsObj = {
+    errors: [],
+  };
+
+  const getError = (errorType) => {
```
AlinaMatuschak, line 19: create new file destructuring with this functions
also, you can create object instead of function:
```
const errorMap = {
  errorType: errorObj
}
```
and get an error from object errorMap.errorType

AlinaMatuschak (approved):

PR #15, opened 2022-10-18
Create case-transforming server

max-kravchenko:
Create a case-transforming server with req JSON representation in the browser


andrii-dmytruk (approved):
Perfect!

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+
+    const errors = [];
+    const errorMessage = {
+      noOriginalText: 'Text to convert is required. '
+     + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      noTargetCase: '"toCase" query param is required. '
+     + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      unsupportedCase: 'This case is not supported. '
+     + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+    };
```
AlinaMatuschak, line 18: 🔥

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+
+    const originalText = req.url.split('?')[0].split('/')[1];
```
AlinaMatuschak, line 20: you can parse url using constructor [URL](https://nodejs.org/api/url.html#constructing-a-url-from-component-parts-and-getting-the-constructed-string). And use pathname from parsed url

AlinaMatuschak (approved):

PR #22, opened 2022-10-18
createServer

NataliyaAn:


src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+
+function createServer() {
+  return http.createServer((req, res) => {
+    const url = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = url.pathname.slice(1);
+    const targetCase = url.searchParams.toCase;
+    const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    let error = false;
+    let messageError = '';
```
sergii-nosachenko, line 12:
```suggestion
    const errors = [];
```
You don't need boolean variable for detecting if you have errors. Simply check if your errors array has length.
And array, not string - here it's best practice, cause you need to response with object { errors: [] }

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+
+function createServer() {
+  return http.createServer((req, res) => {
+    const url = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = url.pathname.slice(1);
+    const targetCase = url.searchParams.toCase;
```
sergii-nosachenko, line 8:
```suggestion
    const targetCase = url.searchParams.get('toCase');
```
You can't get toCase entry directly. You need to use [get method](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+    const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    let error = false;
+    let messageError = '';
+
+    if (!targetCase) {
+      error = true;
```
sergii-nosachenko, line 15:
```suggestion
```

sergii-nosachenko (changes requested):
Hi, Nataliya) Your code has some issues you need to fix.

PR #23, opened 2022-10-18
implemented function case transformer

Mejeezz:


src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+    if (errors.length > 0) {
+      res.statusCode = 400;
+      res.statusText = 'Bad request';
+
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    };
+
+    const convertedText = convertToCase(text, targetCase);
```
AlexanderKolomiiets, line 43:
```suggestion
    const const {
    originalCase,
    convertedText,
  } = convertToCase(text, targetCase);
```

you can destructurize it that way

AlexanderKolomiiets (approved):
👾🐱‍🏍

PR #25, opened 2022-10-18
Solution

DimaKobzar7:


SvyatoslavYatsenko (approved):
GJ

src/createServer.js
```diff
@@ -1,3 +1,68 @@
...
+      res.statusCode = 200;
+
+      response.originalCase = convertToCase(textToConvert, toCase).originalCase;
+
+      response.targetCase = toCase;
+
+      response.originalText = textToConvert;
+
+      response.convertedText = convertToCase(textToConvert, toCase)
+        .convertedText;
```
AlinaMatuschak, line 59: can you create response object here?
and you call convertToCase twice. Can you do it just one time?

src/createServer.js
```diff
@@ -1,3 +1,68 @@
...
+    } else if (!allowedCases.includes(toCase)) {
+      errors.push({
+        message: invalidCaseMessage,
+      });
+    }
+
+    res.setHeader('Content-Type', 'application/json');
+
+    if (errors.length) {
+      res.statusCode = 400;
```
AlinaMatuschak, line 46: Also, you need to return the status test `Bad request`

AlinaMatuschak (changes requested):

src/createServer.js
```diff
@@ -1,3 +1,70 @@
...
+      response.statusText = 'Bad request';
```
AlinaMatuschak, line 47: and your response object should be deleted ) You don’t use it

```suggestion
      res.statusText = 'Bad request';
```

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #35, opened 2022-10-18
Solution

SvyatoslavYatsenko:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+
+    res.statusCode = 200;
+
+    res.end(JSON.stringify({
+      originalCase,
+      targetCase,
+      originalText,
+      convertedText,
+    }));
+  });
```
DimaKobzar7, line 56: You can put it in a condition with errors.length in the else block and then you don't have to write return

DimaKobzar7 (approved):

AlinaMatuschak (approved):

PR #38, opened 2022-10-18
solution

lizapolokhina:


src/createServer.js
```diff
@@ -1,3 +1,77 @@
...
+        ...convertToCase(
+          originalText,
+          targetCase,
+        ),
+        originalText,
+        targetCase,
+      }));
+    }
+
+    res.end();
```
anna-kolmakova, line 69: I believe you don't need this res.end because you already have it above

anna-kolmakova (approved):
Good job 👍

PR #54, opened 2022-10-19
case transformer server

Alej4ndro1:


yungdena (changes requested):
You forgot to set statusCode 200 if there are no errors :)

PR #61, opened 2022-10-20
solution

HusakOleh:


src/createServer.js
```diff
@@ -1,3 +1,51 @@
...
+    }
+
+    res.statusCode = 200;
+
+    res.end(JSON.stringify({
+      originalCase: convertToCase(text, toCase).originalCase,
+      targetCase: toCase,
+      originalText: text,
+      convertedText: convertToCase(text, toCase).convertedText,
+    }));
```
yevhenii-podplietnii, line 47: Better call function only once, and then write result

yevhenii-podplietnii (approved):
Good job!

src/createServer.js
```diff
@@ -1,3 +1,51 @@
...
+  const toCase = normalizedURL.searchParams.get('toCase');
+
+    res.setHeader('Content-type', 'application/json');
+
+    const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+    const errors = [];
+
+    if(!text) {
+      errors.push({
+        message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
dellachaise, line 17: Do not rely on the unknown string, make constants for this or create a special object with all messages. It will be good for future reusability.

dellachaise (approved):
Well done 👍

PR #69, opened 2022-11-04
solution

ihorandrianov:


src/isError.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable */
```
sergii-nosachenko, line 1: Don't disable linter

src/isError.js
```diff
@@ -0,0 +1,30 @@
...
+
+function isError(q, c) {
```
sergii-nosachenko, line 3: Don't use one-char var names. Use proper names.

src/isError.js
```diff
@@ -0,0 +1,30 @@
...
+  }
+  if (!c) {
+    errorObj.errors.push({
+      message:
+        'toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>"',
+    });
+  } else if (!cases.includes(c)) {
+    errorObj.errors.push({
+      message:
+        'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
sergii-nosachenko, line 23: Use template string and `cases` constant to generate list of cases.

src/isError.js
```diff
@@ -0,0 +1,30 @@
...
+      message:
+        'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+  if (!c) {
+    errorObj.errors.push({
+      message:
+        'toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>"',
+    });
+  } else if (!cases.includes(c)) {
```
sergii-nosachenko, line 20: Avoid `else if` branches. Better separate to single if logic.

src/isError.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable */
+
+function isError(q, c) {
+  const errorObj = {
+    errors: [],
+  };
```
sergii-nosachenko, line 6: Why do you use object instead of simple array?

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+      .split('?')
+      .map((a) => a.replace('toCase=', ''));
+    const error = isError(query, caseToTransform);
+
+    if (error.errors.length) {
+      res.writeHead(400, 'Bad request', {
+        'Content-Type': 'application/json',
+      });
+      res.end(JSON.stringify(error));
+    } else {
```
sergii-nosachenko, line 18: Don't use `else`. Better terminate error branch with `return`;

src/createServer.js
```diff
@@ -1,3 +1,40 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { isError } = require('./isError');
+const { convertToCase } = require('./convertToCase');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const [query, caseToTransform] = req.url
```
sergii-nosachenko, line 7: Why not to use `originalText` and `targetCase` as names?

sergii-nosachenko (changes requested):

src/isError.js
```diff
@@ -0,0 +1,30 @@
...
+
+  if (!typeCase) {
+    errors.push({
+      message: 'toCase" query param is required.',
+    });
+  }
+
+  if (!cases.includes(typeCase)) {
+    errors.push({
+      message: `This case is not supported. Available cases: ${cases.reduce(
```
sergii-nosachenko, line 20: Use join method. Now result string is incorrect.

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+const { isError } = require('./isError');
+const { convertToCase } = require('./convertToCase');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const [originalText, targetCase] = req.url
+      .slice(1)
+      .split('?')
+      .map((a) => a.replace('toCase=', ''));
+    const error = isError(originalText, targetCase);
```
sergii-nosachenko, line 11:
```suggestion
    const { errors } = isError(originalText, targetCase);
```

sergii-nosachenko (changes requested):

PR #74, opened 2022-11-20
converToCase

illia-sosnytskyi:


src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normalizeUrl = new URL(req.url, `https://${req.headers.host}`);
+
+    let response = {};
+    const errors = [];
+
+    const textToConvert = normalizeUrl.pathname.slice(1);
+    const toCase = normalizeUrl.searchParams.get('toCase');
```
TheVovchik, line 13: we can name it  to use direct in output object targetCase

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+const availableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normalizeUrl = new URL(req.url, `https://${req.headers.host}`);
+
+    let response = {};
+    const errors = [];
+
+    const textToConvert = normalizeUrl.pathname.slice(1);
```
TheVovchik, line 12: we can name it originalText to use direct as a key and value in output object

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+    const toCase = normalizeUrl.searchParams.get('toCase');
+
+    if (textToConvert.length === 0) {
```
TheVovchik, line 15: will be better to split in separate file logic for errors

TheVovchik (approved):
try to split your logic and avoid monster size functions

in other great job 🔥

src/createServer.js
OleksiiBlezniuk: agree, but it's not critical

src/createServer.js
OleksiiBlezniuk: agree, but it's not critical

src/createServer.js
OleksiiBlezniuk: agree, move this logic to separate function

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+      errors.push({
+        message: 'Text to convert is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
OleksiiBlezniuk, line 18: This message can be moved to constant

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+        message: 'This case is not supported. '
+        + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    if (errors.length > 0) {
+      response.errors = errors;
+      res.statusCode = 404;
+      res.statusText = 'Bad request';
+    } else {
```
OleksiiBlezniuk, line 38: Can we use `end` method and stop function using `return` instead of using `else` block?

OleksiiBlezniuk (changes requested):
GJ! Left few comments

PR #77, opened 2022-11-21
solution was added

Maksym845:


src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+
+function createServer() {
+  return http.createServer((req, res) => {
+    const normalizedUrl = new URL(req.url, `http://${req.headers.host}`);
+
+    const targetCase = normalizedUrl.searchParams.get('toCase');
+    const originalText = normalizedUrl.pathname.slice(1);
+
+    const errors = [];
+    const validCases = ['UPPER', 'SNAKE', 'KEBAB', 'PASCAL', 'CAMEL'];
```
OleksiiBlezniuk, line 13: Let's move it outside function, would be better to move it to `constants` file

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+
+    if (!targetCase) {
+      errors.push('"toCase" query param is required. Correct request is: '
+        + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".');
```
OleksiiBlezniuk, line 17: These error messages also can be moved to constants

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+  return http.createServer((req, res) => {
+    const normalizedUrl = new URL(req.url, `http://${req.headers.host}`);
+
+    const targetCase = normalizedUrl.searchParams.get('toCase');
+    const originalText = normalizedUrl.pathname.slice(1);
+
+    const errors = [];
+    const validCases = ['UPPER', 'SNAKE', 'KEBAB', 'PASCAL', 'CAMEL'];
+
+    if (!targetCase) {
```
OleksiiBlezniuk, line 15: Move this logic to separate function

OleksiiBlezniuk (changes requested):
GJ! Left few comments

PR #79, opened 2022-11-22
implemented solution

ArtemHulkoDev:


src/createServer.js
```diff
@@ -1,3 +1,63 @@
...
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normalizedUrl = new URL(req.url, `http:${req.headers.host}`);
+
+    let response = {};
+    const errors = [];
+    const caseTypeList = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
OleksiiBlezniuk, line 13: Better to move this constant outside function

src/createServer.js
```diff
@@ -1,3 +1,63 @@
...
+      errors.push({
+        message: '"toCase" query param is required. Correct request is: '
+          + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!caseTypeList.includes(toCase)) {
+      errors.push({
+        message: 'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
```
OleksiiBlezniuk, line 35: Better to move this logic to separate function

src/createServer.js
```diff
@@ -1,3 +1,63 @@
...
+
+    if (toCase === null || toCase.length === 0) {
+      errors.push({
+        message: '"toCase" query param is required. Correct request is: '
+          + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!caseTypeList.includes(toCase)) {
+      errors.push({
+        message: 'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
OleksiiBlezniuk, line 33: In case if list of this cases will be changed, this message will include wrong cases, so better to use values from `caseTypeList`

OleksiiBlezniuk (changes requested):
GJ! Left few comments to fix

PR #82, opened 2022-11-25
case transformer server was created

iPanda177:


src/createServer.js
```diff
@@ -1,3 +1,74 @@
...
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const normaliseUrl = new URL(req.url, `http://${req.headers.host}`);
+    const { pathname, searchParams } = normaliseUrl;
+    const originalText = pathname.slice(1);
+    const params = new URLSearchParams(searchParams);
+    const targetCase = params.get('toCase');
+    const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
OleksiiBlezniuk, line 22: Let's move it outside this function
Would be better to move constants to separate file

OleksiiBlezniuk (changes requested):
Good job! Left few small comments

src/createServer.js
```diff
@@ -1,3 +1,55 @@
...
+const { convertToCase } = require('./convertToCase');
+const { checkValidateRequest } = require('./validateRequest');
+
+const errorsTypes = {
+  notOriginalText: 'Text to convert is required.'
+  + ' Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  notTargetCase: '"toCase" query param is required.'
+  + ' Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  wrongTargetCase: 'This case is not supported.'
+  + ' Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
OleksiiBlezniuk, line 11: Better to use values from cases list, then you'll be able to just list and this string will be updated

OleksiiBlezniuk (approved):

PR #84, opened 2022-11-27
create server, add logic (find errors and convert text)

SerhiiRomaniukWD:


src/createServer.js
```diff
@@ -1,3 +1,60 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
```
OleksiiBlezniuk, line 5: not critical, but you could return result of this method without saving to constant

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+    const normalizedUrl = new URL(req.url, `http://${req.headers.host}`);
+
+    const convertCases = [
+      'SNAKE',
+      'KEBAB',
+      'CAMEL',
+      'PASCAL',
+      'UPPER',
+    ];
```
OleksiiBlezniuk, line 14: Let's move this constant outside function, for example to `constants` file

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+      errors.push({
+        message: '"toCase" query param is required.'
+          + ' Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!convertCases.includes(params.toCase)) {
+      errors.push({
+        message: 'This case is not supported.'
+          + ' Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
```
OleksiiBlezniuk, line 39: Better to move this logic to separate function, errors also can be moved to constants

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+
+      response.originalCase = result.originalCase;
+      response.targetCase = params.toCase;
+      response.originalText = textToConvert;
+      response.convertedText = result.convertedText;
+    }
+
+    res.setHeader('Content-Type', 'application/json');
+
+    res.end(JSON.stringify(response));
```
OleksiiBlezniuk, line 54: we can use `end` method inside `if` block, and stop function using return instead of `else` block

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,29 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <JSCodeStyleSettings version="0">
+      <option name="FORCE_SEMICOLON_STYLE" value="true" />
```
OleksiiBlezniuk, line 4: I guess this files shouldn't be here 🙂

OleksiiBlezniuk (changes requested):
GJ! Left few comments to fix

PR #96, opened 2022-12-28
First try

Esceype:


src/createServer.js
```diff
@@ -1,3 +1,68 @@
...
+    }
+
+    if (errors.length > 0) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
```
denys-danyliuk, line 45: would be better to create function for this and move it to file with errors

denys-danyliuk (approved):

PR #98, opened 2022-12-28
Solution

Illia-Skladnik:


src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+      };
+
+      res.setHeader('Content-Type', 'application/json');
+
+      res.statusCode = 400;
+      res.statusText = 'Bad request';
+      res.end(JSON.stringify(errorResponse));
+
+      return;
+    }
```
Svitlana-Yudina, line 46:
```suggestion
    if (errors.length > 0) {
      res.setHeader('Content-Type', 'application/json');

      res.statusCode = 400;
      res.statusText = 'Bad request';
      res.end(JSON.stringify({ errors }));

      return;
    }
```
It is not necessary to create a separate variable))

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+      convertedText: resultData.convertedText,
+      originalCase: resultData.originalCase,
+      originalText: text,
+      targetCase: toCase,
+    };
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = 200;
+    res.write(JSON.stringify(result));
+    res.end();
```
Svitlana-Yudina, line 59:
```suggestion
    res.end(JSON.stringify(result));
```
I think, you don't need `res.write` just `res.end()` ))

Svitlana-Yudina (approved):
Great Job!)
You can move a validation to a separate function. But it's up to you))

src/createServer.js
```diff
@@ -1,3 +1,38 @@
...
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { isCodeValid } = require('./modules/isCodeValid');
+const http = require('http');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const [text, toCase] = getQueryInfo(req);
+    const errors = isCodeValid(text, toCase);
+
+    if (errors.length > 0) {
```
mykola-koval, line 11:
```suggestion
    if (errors.length) {
```

mykola-koval, line 11:
```suggestion
    if (errors.length) {
```

mykola-koval: btw you can omit this block at all by placing in after your `else`

src/modules/getQueryInfo.js
```diff
@@ -0,0 +1,12 @@
+function getQueryInfo(req) {
+  const partsUrl = req.url.split('?');
+  const params = new URLSearchParams(partsUrl[1]);
+  const toCase = params.get('toCase');
+  const text = partsUrl[0].slice(1);
```
mykola-koval, line 5: you can also use new URL constructor or this logic to possibly make it shorter

src/modules/isCodeValid.js
```diff
@@ -0,0 +1,31 @@
+function isCodeValid(text, toCase) {
+  const errors = [];
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+  if (text.length === 0) {
+    errors.push({
+      // eslint-disable-next-line max-len
```
mykola-koval, line 7: you can use string concatenation here instead of disabling eslint

mykola-koval (approved):

PR #109, opened 2022-12-29
added solution

kryvets-vadym:


src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+    const params = new URLSearchParams(queryParams);
+    const caseParams = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+    const toCase = params.get('toCase');
+
+    const errors = [];
+
+    if (text.length === 0) {
+      errors.push({
+        // eslint-disable-next-line max-len
+        message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
denys-danyliuk, line 17: create object where you would store your error messages, also create another file for this

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+    }
+
+    if (toCase && !caseParams.includes(toCase.toUpperCase())) {
+      errors.push({
+        // eslint-disable-next-line max-len
+        message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    if (errors.length > 0) {
```
denys-danyliuk, line 35: would be better to create function for this and move to another file

denys-danyliuk (approved):

PR #113, opened 2022-12-29
Case transformer server

referee1988:


Artem66 (approved):
GJ!

mykola-koval (changes requested):
let's move the functions into separate files

src/getErrors.js
```diff
@@ -0,0 +1,34 @@
...
+    errors.push({ message: ErrorMessages.noText });
+  }
+
+  if (!toCase) {
+    errors.push({ message: ErrorMessages.noCase });
+  }
+
+  if (toCase && !cases.includes(toCase)) {
+    errors.push({ message: ErrorMessages.wrongCase });
+  }
```
illiarerun, line 27: would be actually great to use `else if` here

PR #119, opened 2023-01-03
Solution

vasylevskaya:


src/convertToCase/validate.js
```diff
@@ -0,0 +1,20 @@
...
+  let errors = []
+  const VALID_CASES = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER']
+
+  if (!textToConvert?.length) {
+    errors.push("Text to convert is required. Correct request is: /<TEXT_TO_CONVERT>?toCase=<CASE_NAME>")
+  }
+
+  if (!caseToConvert?.length) {
+    errors.push("'toCase' query param is required. Correct request is: /<TEXT_TO_CONVERT>?toCase=<CASE_NAME>")
+  } else if (!VALID_CASES.includes(caseToConvert?.toUpperCase())) {
```
witflash, line 11: You can use syntax like this:
```
if (caseToConvert && !VALID_CASES.includes(caseToConvert.toUpperCase())) {
  errors.push("This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER")
}
```

So you will have more readable code and will not need to use `else if`.

src/convertToCase/validate.js
```diff
@@ -0,0 +1,20 @@
+function validate(textToConvert, caseToConvert) {
+  let errors = []
+  const VALID_CASES = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER']
+
+  if (!textToConvert?.length) {
```
witflash, line 5: You can check more simply:
```
if (!textToConvert)
```

src/convertToCase/validate.js
```diff
@@ -0,0 +1,20 @@
...
+    errors.push("Text to convert is required. Correct request is: /<TEXT_TO_CONVERT>?toCase=<CASE_NAME>")
+  }
+
+  if (!caseToConvert?.length) {
```
witflash, line 9: And here too

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+    res.statusCode = 200
+    res.statusMessage = 'OK'
+    console.log(result)
+  } else {
+    res.statusCode = 400
+    res.statusMessage = 'Bad request'
+    console.log(errors)
+  }
+
+  res.end()
```
witflash, line 28: What about requirements?

```
Response body should be the next JSON:

{
  "originalCase": "CASE_NAME",
  "targetCase": "CASE_NAME",
  "originalText": "ORIGINAL_TEXT",
  "convertedText": "CONVERTED_TEXT"
}

```

witflash (changes requested):
Good work.
But carefully check the task requirements.

witflash (approved):
Much better! Well done 👍

PR #120, opened 2023-01-15
feat(convert): added the ability to convert text by case

Vladyslav317:


src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+        message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!cases.includes(toCase)) {
+      errors.push({
+        // eslint-disable-next-line max-len
+        message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    const data = convertToCase(text, toCase);
```
marina-tilniak, line 35: You should call convertToCase only if there is no errors

marina-tilniak (approved):

PR #125, opened 2023-01-31
Solution

shyptia:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const toCaseValues = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const respondErrors = [];
+
+    const requestParts = req.url.split('?');
+    const textToConvert = requestParts[0].slice(1);
```
solaryasha, line 11: I guess, array destructuring would be better for code reader
```
const [endpoint, queryParams] = req.url.split('?');

File: src/createServer.js [hunk 2]
```diff
@@ -1,3 +1,61 @@
...
+          + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!toCaseValues.includes(toCase)) {
+      respondErrors.push({
+        message: 'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    if (respondErrors.length !== 0) {
```
solaryasha, line 36:
```suggestion
    if (respondErrors.length) {
```

src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+      res.statusMessage = 'Bad request';
+
+      res.end(JSON.stringify({
+        errors: respondErrors,
+      }));
+    } else {
+      const result = convertToCase(textToConvert, toCase);
+
+      res.statusCode = 200;
+      req.statusMessage = 'OK';
```
solaryasha, line 47: It should be response message

solaryasha (changes requested):

src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+      respondErrors.push({
+        message: '"toCase" query param is required. Correct request is: '
+          + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!toCaseValues.includes(toCase)) {
+      respondErrors.push({
+        message: 'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
```
solaryasha, line 34: Maybe, let's move this data validation to separate function ?

src/createServer.js
```diff
@@ -1,3 +1,41 @@
...
+    const textToConvert = endpoint.slice(1);
+    const params = new URLSearchParams(queryParams);
+    const toCase = params.get('toCase');
+
+    res.setHeader('Content-Type', 'application/json');
+
+    const respondErrors = checkForErrors(textToConvert, toCase);
+
+    if (respondErrors.length) {
+      res.statusCode = 404;
```
solaryasha, line 17: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

PR #129, opened 2023-02-13
node case transformer server solution

anton-kulchytskyi:


src/createServer.js
```diff
@@ -1,3 +1,41 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/index');
```
witflash, line 2: You can do import directly from `require('./convertToCase')`

witflash (approved):
Well done 👍 
I like it!

PR #139, opened 2023-02-16
solution

Fynkzia:


nikachu404 (approved):
GJ! 🚀

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+      res.statusMessage = 'OK';
+
+      res.end(JSON.stringify({
+        originalCase,
+        targetCase,
+        originalText,
+        convertedText,
+      }));
+    } else {
+      res.statusCode = 404;
```
solaryasha, line 30:
```suggestion
      res.statusCode = 400;
```
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

solaryasha: Let's return proper status code

solaryasha (approved):

PR #146, opened 2023-02-17
sol

KrisPlanka:


src/createServer.js
```diff
@@ -1,3 +1,74 @@
...
+      ArrOfErrors.push({
+        message: '"toCase" query param is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!cases.includes(toCase)) {
+      ArrOfErrors.push({
+        message: 'This case is not supported. '
+        + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
```
oleh-yablunovskyi, line 43: Maybe it's better to move URL validation to separate function.

oleh-yablunovskyi (approved):
Good 👍

src/createServer.js
```diff
@@ -1,3 +1,74 @@
...
+
+    const cases = [
+      'SNAKE',
+      'KEBAB',
+      'CAMEL',
+      'PASCAL',
+      'UPPER',
+    ];
+
+    const ArrOfErrors = [];
```
ivan-kadykalo, line 22: Don't start names from upper case if it is not Component, Class or Enum
```suggestion
    const errorMessages = [];
```

src/createServer.js
ivan-kadykalo: Completely agree, move URL validation logic to the separate function

ivan-kadykalo (changes requested):
You are almost there 💪

src/createServer.js
KrisPlanka: fixed, thank you!

ivan-kadykalo (approved):
Good job 🔥

PR #158, opened 2023-02-19
solution

olena-yanovska:


src/createServer.js
```diff
@@ -1,3 +1,44 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { valigateData } = require('./validateData');
```
daryna-zinchenko, line 2: should it be vali(d)ateData?

daryna-zinchenko (approved):
Great job!

src/createServer.js
ivan-kadykalo: Yep 🥲, should be vali(d)ate

olena-yanovska: Oh, really, thank you!

PR #175, opened 2023-03-01
solution

igor-kashchenko:


src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+    const textToConvert = normalizedURL.pathname.slice(1);
+    const toCase = normalizedURL.searchParams.get('toCase');
+    const errors = validator(textToConvert, toCase);
+
+    if (errors.errors.length) {
+      req.statusCode = 404;
+      req.statusMessage = 'Bad request';
+      res.end(JSON.stringify(errors));
+    } else {
+      res.statusCode = 200;
```
nzrii-kysel, line 19: 200 is default
```suggestion
```

src/validator.js
```diff
@@ -0,0 +1,20 @@
...
+const validCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+function validator(text, toCase) {
+  return {
+    errors: [
+      ...(!text ? [{ message: errors.textIsMissing }] : []),
+      ...(!toCase ? [{ message: errors.toCaseIsMissing }] : []),
+      ...(!validCases.includes(toCase) && toCase ? [{ message: errors.toCaseIsNotSupported }] : []),
+    ],
+  };
```
nzrii-kysel, line 17: this feels like something from codewars. not sure if the code like that has a place in a real project

nzrii-kysel (approved):
well done 👍

src/validator.js
denys-cheporniuk: agree, It's better to make this code more readable

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #176, opened 2023-03-06
case transform solution

PolinaSukhorukova:


src/createServer.js
```diff
@@ -1,3 +1,35 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
```
anna-inozemets, line 2: It isn't so iportant, but I guess that it will work with shorter path like ('./convertToCase')

anna-inozemets (approved):
Clear and good solution!)

PR #177, opened 2023-03-07
Initial commit of creating server

anna-inozemets:


src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const textToConvert = normalizedURL.pathname.slice(1);
+    const toCase = normalizedURL.searchParams.get('toCase');
+
+    const errors = handleErrors(textToConvert, toCase);
+
+    if (!errors.errors.length) {
+      const convertedText = convertToCase(textToConvert, toCase);
+
+      req.statusMessage = 'OK';
```
PolinaSukhorukova, line 18: I think, that statusMessage for statusCode = 200 is Ok by default.

PolinaSukhorukova (approved):
GJ 👍

src/createServer.js
denys-cheporniuk: yeah, OK should by default

denys-cheporniuk (approved):

PR #178, opened 2023-03-07
Solution

andrieieva-yuliia:


src/validateErrors.js
```diff
@@ -0,0 +1,34 @@
...
+  {
+    message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  },
+  {
+    message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  },
+  {
+    message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+  },
+];
```
Bogdan-Kutasevich, line 12: Maybe better to use object for errors or even consts, because in code below hard to understand what error do you push.

src/createServer.js
```diff
@@ -1,3 +1,33 @@
...
+const { normalizeURL } = require('./normalizeUrl');
+const { validateErrors } = require('./validateErrors');
+const { sendResponse } = require('./sendResponse');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const paramsToConvert = normalizeURL(req.url);
+    const errors = validateErrors(...paramsToConvert);
+
+    if (errors.length > 0) {
```
Bogdan-Kutasevich, line 12: if(errors.length)

Bogdan-Kutasevich (approved):
Nice work)
check comment below.

src/validateErrors.js
alexandra-protyanova: agree, not clear at all, high probability of mistake

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Good job 👍

PR #179, opened 2023-03-07
Task done. Have a good time!)

Bogdan-Kutasevich:


src/createServer.js
```diff
@@ -1,3 +1,44 @@
...
+      res.end();
+    } else {
+      const response = {
+        errors,
+      };
+
+      res.setHeader('Content-Type', 'application/json');
+      res.statusCode = 400;
+      res.write(JSON.stringify(response));
+      res.end();
```
GarryRocksman, line 35: the same logic here with sending response, think how you can optimize it

src/getErrors.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable max-len */
+const emptyTextError = 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+const emptyCaseError = '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+const NotSupportedCaseError = 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.';
```
GarryRocksman, line 4: nice sollution with variables here

src/createServer.js
```diff
@@ -1,3 +1,44 @@
...
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normilizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const textToTransform = normilizedURL.pathname.slice(1);
+    const typeToTransform = normilizedURL.searchParams.get('toCase');
+    const errors = getErrors(textToTransform, typeToTransform);
+
+    if (!errors.length) {
+      const transformedData = convertToCase(textToTransform, typeToTransform);
```
GarryRocksman, line 16: maybe here also you should use try-catch, to avoid some unexpectable errors

GarryRocksman (approved):
GJ 🔥, just go through my comments maybe you find something helpful

denys-cheporniuk (approved):

PR #180, opened 2023-03-07
Case transformer server - completed

GarryRocksman:
- Have a nice Day 🔥

### DIFF
--- a/src/modules/urlParamsPreValidation.js
+++ b/src/modules/urlParamsPreValidation.js
@@ -1,0 +1,33 @@
+'use strict';
+
+const urlParamsPreValidation = (textToConvert, targetCase) => {
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+  const caughtErrors = {
+    errors: [],
+  };
+
+  const validCase = cases.find(c => c === targetCase);
+
+  if (!targetCase) {
+    caughtErrors.errors.push({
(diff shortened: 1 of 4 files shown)


svitjojo (changes requested):
I have just some questions:
 1. Why do you check URL for favicon?
 2. In urlParamsPreValidation you used the if-else statement and after separate if, I think it's better to use one code style (only if else or if)
 3. Try catch? Nice 👍 
 
 Mentors don't like if-else, so fix it 😉

GarryRocksman:
`Why do you check URL for favicon?`
Because the browser sends a request for it , it's the default request, it's annoying, so I caught it ))))

src/modules/urlParamsPreValidation.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const urlParamsPreValidation = (textToConvert, targetCase) => {
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+  const caughtErrors = {
+    errors: [],
+  };
```
alexandra-protyanova, line 7: I think it is better to return just array

alexandra-protyanova (approved):

src/modules/urlParamsPreValidation.js
GarryRocksman: fixed

PR #181, opened 2023-03-07
Solution

Avichez:


src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+    if (!errors.length) {
+      const convertedData = convertToCase(originalText, targetCase);
+
+      res.data = JSON.stringify({
+        originalText,
+        targetCase,
+        ...convertedData,
+      });
+    } else {
+      res.statusCode = 404;
```
tania-boiar, line 25: Needed here 400 error.
404 is a 'Not found' error.

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+      res.statusMessage = 'Bad request';
+      res.data = JSON.stringify({ errors });
+    }
+
+    res.setHeader('Content-Type', 'application/json');
```
tania-boiar, line 30: I think you need to move it up before the ` if/else` statement.

tania-boiar (approved):
Well done! 🔥

src/createServer.js
Avichez: Oops, my bad 🥲

src/createServer.js
Avichez: If I'm not mistaken the final response is  `res.end`  and `res.setHeader` exactly above it.

alexandra-protyanova (approved):
Good job 🚀

PR #183, opened 2023-03-07
Solution

tania-boiar:


src/createServer.js
```diff
@@ -1,3 +1,33 @@
...
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const toCase = normalizedURL.searchParams.get('toCase');
+    const pathname = normalizedURL.pathname.slice(1);
+    const errors = validate(pathname, toCase);
+
+    if (errors.length) {
+      req.statusCode = 400;
+      req.statusMessage = 'Bad request';
+      res.end(JSON.stringify({ errors }));
+    } else {
```
BrandonSunrise, line 18: Shouldnt we hande default "OK' case?

src/validate.js
```diff
@@ -0,0 +1,27 @@
...
+      + ' Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  } else if (!(validCases.includes(toCase))) {
+    errors.push({
+      message: 'This case is not supported.'
+      + ' Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+    });
+  }
+
+  return errors;
```
BrandonSunrise, line 24: As for me - i would move validCases and error messages to different file . So if our valid cases will change we won't need to rewrite the logic files.

src/createServer.js
```diff
@@ -1,3 +1,33 @@
...
+      const convertedPath = convertToCase(pathname, toCase);
```
BrandonSunrise, line 19: You can make logic of this object more simple - you can use either spread operator or destructurization in function call

BrandonSunrise (approved):
Thats a nice code! But as for my opinion i would make some tiny changes.

src/validate.js
alexandra-protyanova: I would also create an object for possible errors

src/createServer.js
alexandra-protyanova: agree, you can destructure needed properties from object

src/createServer.js
alexandra-protyanova: 👍

alexandra-protyanova (changes requested):

src/createServer.js
tania-boiar: Don't we already have `200 Ok` by default?

alexandra-protyanova: Sorry, you are right it is default

alexandra-protyanova (approved):
Well done 👍

PR #187, opened 2023-03-07
solved task

Serghii29:


src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+        message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    if (errors.length > 0) {
+      res.writeHead(400, { 'Content-type': 'application/json' });
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
```
OlehSharha, line 38: i think its better to create separate file for this, its only my opinion)

src/createServer.js
```diff
@@ -1,3 +1,56 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const [text, queryString] = req.url.split('?');
```
OlehSharha, line 7: Interesting solution i like it:)

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+      });
+    }
+
+    if (!toCase) {
+      errors.push({
+        message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
OlehSharha, line 25: i think it can be better if you write it higher, all const in one place

OlehSharha (approved):
Nice job)))

alexandra-protyanova (changes requested):
Split your logic into separate files

src/catchError.js
```diff
@@ -0,0 +1,26 @@
...
+  { message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.' },
+];
+
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+export const catchError = (text, toCase) => {
+  const errors = [];
+
+  if (!text || text === '/') {
+    errors.push(massageError[0]);
```
alexandra-protyanova, line 14: not a good idea to take elements by indexes, let us make `massageError` object, not array

src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
+const { convertToCase } = require('./convertToCase');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const [text, queryString] = req.url.split('?');
+    const params = new URLSearchParams(queryString);
+    const toCase = params.get('toCase');
+    const errors = catchError(text, toCase);
+
+    if (errors.length > 0) {
```
alexandra-protyanova, line 13:
```suggestion
    if (errors.length) {
```

alexandra-protyanova (changes requested):

PR #189, opened 2023-03-07
Solution

max-rozzhalovets:


OlhaYevstifieieva (changes requested):
This code is awkward to read. You should split it into several files. At least take out separately handling of errors. Better yet, create an object from the error options, and call messages by key.

denys-cheporniuk (approved):

PR #197, opened 2023-03-08
add server

Anton-Zhytonbaiev:


src/createServer.js
```diff
@@ -1,3 +1,36 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
```
OleksandrProkop, line 2: enough ./convertToCase

src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
+    res.setHeader('Content-Type', 'application/json');
+
+    const normalizeURL = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = normalizeURL.pathname.slice(1);
+    const targetCase = normalizeURL.searchParams.get('toCase');
+    const errors = validation(originalText, targetCase);
+
+    if (errors.errors.length) {
+      req.statusCode = 404;
+      req.statusMessage = 'Bad request';
```
OleksandrProkop, line 16: i think we should respond with 400 status, not request

OleksandrProkop (approved):
gj

src/validation.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+  }
+
+  if (!supportedCases.includes(targetCase) && targetCase) {
+    allErrors.push({
+      message: errorMessage.notSupportedCase,
+    });
+  }
+
+  return { errors: allErrors };
```
alexandra-protyanova, line 25: I think it is better to return array of errors

src/createServer.js
alexandra-protyanova: agree

alexandra-protyanova (changes requested):

PR #201, opened 2023-03-08
solution

vladyslav78292:


src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+const { handleError } = require('./errorMessages');
+
+function createServer() {
+  const getServer = http.createServer((request, response) => {
+    response.setHeader('Content-type', 'application/json');
+
+    const updatedUrl = new URL(request.url, `http://${request.headers.host}`);
+    const textToConvert = updatedUrl.pathname.slice(1) || '';
+    const caseToConvert = updatedUrl.searchParams.get('toCase') || '';
+    const possibleError = handleError(textToConvert, caseToConvert);
```
GarryRocksman, line 12: Just suggestion, it s better to use more undestandeble naming here? like handleErrors.

src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+        originalText: textToConvert,
+        targetCase: caseToConvert,
+      }));
+    }
+  });
+
+  return getServer;
+}
+
+module.exports = { createServer };
```
GarryRocksman, line 31:
```suggestion
module.exports = { 
createServer
 };
```

src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+const { convertToCase } = require('./convertToCase');
+const { handleError } = require('./errorMessages');
+
+function createServer() {
+  const getServer = http.createServer((request, response) => {
+    response.setHeader('Content-type', 'application/json');
+
+    const updatedUrl = new URL(request.url, `http://${request.headers.host}`);
+    const textToConvert = updatedUrl.pathname.slice(1) || '';
+    const caseToConvert = updatedUrl.searchParams.get('toCase') || '';
```
GarryRocksman, line 11: If am not mistaken there is no need || because in case of apsent params, there are already empty strings

src/errorMessages.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable max-len */
+const missingText = 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+const caseMissing = '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+const wrongCaseValue = 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.';
```
GarryRocksman, line 4: nice sollution with messages variables

src/errorMessages.js
```diff
@@ -0,0 +1,26 @@
...
+
+  if (!toCase) {
+    errors.push({ message: caseMissing });
+  }
+
+  if (!availableCases.includes(toCase) && toCase) {
+    errors.push({ message: wrongCaseValue });
+  };
+
+  return { errors };
```
GarryRocksman, line 23: on the mentors review Sasha wrote me that its better to return an array here

src/errorMessages.js
```diff
@@ -0,0 +1,26 @@
...
+};
+
+module.exports = { handleError };
```
GarryRocksman, line 26:
```suggestion
module.exports = { 
handleError 
};
```

GarryRocksman (approved):
GJ 🔥, Just look at the comment, maybe you will find something useful

ArturErmolenko (approved):

PR #203, opened 2023-03-08
implemented case_transformer

andrii-kurash:


src/checkErrors.js
```diff
@@ -0,0 +1,34 @@
...
+  {
+    message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+  },
+];
+
+const checkErrors = (url, toCase) => {
+  const errorMessage = [];
+
+  if (!url.length) {
+    errorMessage.push(errors[0]);
```
Anna-Tkachenko, line 18: it's not stable solution with getting an errors by index
will be better to create separate constants for each of them

Anna-Tkachenko (approved):

PR #221, opened 2023-04-12
Create "Case Transformer Server"

rostyslav-horytskyi:


src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
+    const toCase = searchParams.get('toCase');
+    const errors = validate(text, toCase);
+
+    res.setHeader('Content-Type', 'application/json');
+
+    if (errors.length) {
+      res.writeHead(400, 'Bad request');
+      // eslint-disable-next-line no-console
+      errors.forEach(error => console.error(error));
+      res.end('Wooops! Something went wrong 😢');
```
nazarmatsevych, line 21: It's recommended to return the errors array in the response body instead of a generic error message.

src/validate.js
```diff
@@ -0,0 +1,35 @@
+'use strict';
+
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
nazarmatsevych, line 3: Move cases to separate file

src/validate.js
```diff
@@ -0,0 +1,35 @@
...
+
+function validate(text, toCase) {
```
nazarmatsevych, line 5: The function name 'validate' is too generic, consider renaming it to 'validateTextConversion' or smth else for better clarity.

nazarmatsevych (approved):
Good work,
left some minor comments

PR #223, opened 2023-04-14
Server was created

lilia-obushenko:


src/checkUrl.js
```diff
@@ -0,0 +1,36 @@
...
+  'UPPER',
+];
+
+const checkUrl = (text, toCase) => {
+  const errors = [];
+
+  if (!text) {
+    errors.push({
+      message: 'Text to convert is required. '
+      + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
Oleksandr-kopaevich, line 15: It's good practice to use `template literals` like this
```suggestion
      message: `Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".`,
```
It doesn't require to escape special characters (text inside will be as it is including spacing). Also you can include variables inside.

src/createServer.js
```diff
@@ -1,3 +1,47 @@
...
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const [text, query] = req.url.split('?');
+    const { originalText, targetCase } = getParams(text, query);
+
+    const errors = checkUrl(originalText, targetCase);
+
+    if (errors.length) {
+      res.statusCode = 404;
```
Oleksandr-kopaevich, line 16: It's the wrong code. 404 - means that such endpoint wasn't found or some entity in the database, or requested case if talking closer to this example. All your errors are related to incorrect params that someone has sent you. So you must return statusCode: 400 for a Bad Request.

src/checkUrl.js
```diff
@@ -0,0 +1,36 @@
...
+      message: '"toCase" query param is required. '
+      + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!cases.includes(toCase) && toCase) {
+    errors.push({
+      message: 'This case is not supported. '
+      + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+    });
```
Oleksandr-kopaevich, line 30: You have the variable `cases`, use values from it to have always up to date errors
```suggestion
    errors.push({
      message: `This case is not supported. Available cases: ${cases.join(', ')}.`,
    });
```

Oleksandr-kopaevich (approved):
Hey @lilia-obushenko, well done, all tests are passed. I've just left a few comments about how you can improve your future code. Keep going a young conqueror of Nodejs :)

lilia-obushenko:
> Hey @lilia-obushenko, well done, all tests are passed. I've just left a few comments about how you can improve your future code. Keep going a young conqueror of Nodejs :)

PR #225, opened 2023-04-21
Solution

timoxa2t:


src/createServer.js
```diff
@@ -1,3 +1,52 @@
...
+      errors.push({ message: 'Text to convert is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
+    }
+
+    if (targetCase === null) {
+      errors.push({ message: '"toCase" query param is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
+    } else if (!cases.includes(targetCase)) {
+      errors.push({ message: 'This case is not supported. '
+        + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.' });
```
Oleksandr-kopaevich, line 23: It's good practice to use template literals like this. It doesn't require to escape special characters (text inside will be as it is including spacing). Also, you can include variables inside. You have the variable cases, and use values from it to have always up-to-date errors

```suggestion
      errors.push({ message: `This case is not supported. Available cases: ${cases.join(', ')}.`});
```

src/createServer.js
```diff
@@ -1,3 +1,52 @@
...
+    }
+
+    if (errors.length) {
+      res.statusCode = 400;
+      res.statusText = 'Bad request';
+      res.setHeader('Content-Type', 'application/json');
```
Oleksandr-kopaevich, line 29: Your server returns the same response all the time, so set it just once at the start of the request handling.

Oleksandr-kopaevich (approved):
Hey @timoxa2t, all tests have passed, so I'm approving. Just check some comments about how the code could be improved.

PR #228, opened 2023-04-28
case transformer done

rashyd-hasratov:


src/URLValidator.js
```diff
@@ -0,0 +1,31 @@
...
+    errors.push(noTextError);
+  }
+
+  if (!toCase) {
+    errors.push(noToCaseError);
+  } else if (!toCases.includes(toCase)) {
+    errors.push(invalidToCaseError);
+  }
+
+  return errors;
```
KuharchukDmytro, line 28: better to write separate if blocks
```suggestion
  if (!toCase) {
    errors.push(noToCaseError);
  }
  
  if (!toCases.includes(toCase) && toCase) {
    errors.push(invalidToCaseError);
  }

  return errors;
```

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+      const responseBody = JSON.stringify({
+        originalCase,
+        targetCase: toCase,
+        originalText: textToConvert,
+        convertedText,
+      });
+
+      res.end(responseBody);
+    }
+  });
```
KuharchukDmytro, line 35:
```suggestion
    if (errors.length) {
      res.statusCode = 400;
      res.statusMessage = 'Bad request';
      res.end(JSON.stringify({ errors }));
      
      return;
    }
    
    const {
      originalCase,
      convertedText,
    } = convertToCase(textToConvert, toCase);

    const responseBody = JSON.stringify({
      originalCase,
      targetCase: toCase,
      originalText: textToConvert,
      convertedText,
    });

    res.end(responseBody);
  });
```

KuharchukDmytro (approved):
well done, only one remark about if statement nesting, better to avoid it, but otherwise everything is good

main.js
```diff
@@ -2,6 +2,5 @@ const { createServer } = require('./src/createServer');
 
 createServer()
   .listen(5700, () => {
-    // eslint-disable-next-line no-console
-    console.log('Server started! 🚀');
+    global.console.log('Server started! 🚀');
```
juljaVysotska, line 5: Remove the 'global.' prefix and revert to using 'console.log' directly. Instead of disabling 'no-console' rule for this line, consider using a dedicated logger or using custom wrapper for `console.log`

src/constants.js
```diff
@@ -0,0 +1,17 @@
+/* eslint-disable max-len */
```
juljaVysotska, line 1: Avoid using 'eslint-disable' for specific rules. Consider adjusting the max-len rule in your ESLint configuration to an appropriate value.

src/constants.js
```diff
@@ -0,0 +1,17 @@
...
+
+const noTextError = {
```
juljaVysotska, line 3: Here you have 3 variables with same structure for text data. Will be better if you create an object, add in this object text props and export only object. Because in future when you want to add some new text data you have to create a new variable, export this variable and import in another file. Many different steps. But if you use an object, you add new prop and export object only one time. You don't need to import smth else because all info in your object.

juljaVysotska (changes requested):
Good work! Fix some correction and will be perfect!

Anton-Tretiak (approved):

OleksiiBlezniuk (approved):

juljaVysotska (approved):
GJ☺️

PR #229, opened 2023-04-28
case-transformer-server was implement

KuharchukDmytro:


src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+const { getErrors } = require('./getErrors');
+
+const createServer = () => {
+  return http.createServer((request, response) => {
+    const [text, searchParams] = request.url.slice(1).split('?');
+    const params = new URLSearchParams(searchParams);
+    const toCase = params.get('toCase');
+    const errors = getErrors(text, toCase);
```
Oleksandr-kopaevich, line 10: I would use validate in this context. get it's when you just retrieve information, but here you are checking request params, so name can be `validateRequest` or `validateRequestParams`

Oleksandr-kopaevich (approved):
Hey @KuharchukDmytro, well done. All tests have passed.

src/getErrors.js
```diff
@@ -0,0 +1,31 @@
...
+    errors.push({
+      message: '"toCase" query param is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!availableCases.includes(toCase) && toCase) {
+    errors.push({
+      message: 'This case is not supported. '
+        + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
Oleksandr-kopaevich, line 22: It's good practice to use template literals like this. It doesn't require to escape special characters (text inside will be as it is including spacing). Also, you can include variables inside. You have the variable cases, and use values from it to have always up-to-date errors

```suggestion
      message: `This case is not supported. Available cases: ${availableCases.join(', ')}.`,
```

PR #233, opened 2023-05-02
case-transformer-server created

vladyslav-mosor:


milenadiamonds1 (approved):
GJ!

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normalizedUrl = new URL(req.url, `http://${req.headers.host}`);
+    const { pathname, searchParams } = normalizedUrl;
+
+    const text = pathname.slice(1);
+    const toCase = searchParams.get('toCase');
+
+    const errors = validateUrl(text, toCase);
```
juljaVysotska, line 13: Consider renaming 'errors' to 'validationErrors' to provide better context.

src/errorMessages.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable max-len */
+const errorMessages = {
```
juljaVysotska, line 2: Since there is no logic here, it would be possible to use JSON instead of JS

src/validateUrl.js
```diff
@@ -0,0 +1,33 @@
...
+    );
+  }
+
+  if (!toCase) {
+    errors.push(
+      {
+        message: errorMessages.MissingCase,
+      },
+    );
+  } else if (!caseNames.includes(toCase)) {
```
juljaVysotska, line 22: This if statement we can separate too. Like, we have 3 different if statements for different situations.

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+      res.statusMessage = 'Bad Request';
+      res.setHeader('Content-Type', 'application/json');
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const convertedData = convertToCase(text, toCase);
+
+    res.statusCode = 200;
```
juljaVysotska, line 26: You have duplicated code here and on 16-19 lines. We can create a function with arguments to avoid duplication. That function just receives status code, message and data for JSON.stringify into arguments. It will be easily to use and reuse this function in future.

src/createServer.js
```diff
@@ -1,3 +1,40 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { validateUrl } = require('./validateUrl');
+const { convertToCase } = require('./convertToCase');
```
juljaVysotska, line 3: Not added this file :(

src/validateUrl.js
```diff
@@ -0,0 +1,33 @@
...
+    errors.push(
+      {
+        message: errorMessages.WrongCase,
+      },
+    );
+  }
+
+  return errors;
```
juljaVysotska, line 30: If this function validates data, maybe makes sense to throw a new error if validation failed? This function can return nothing, BUT throw error if array of errors is not empty. This behaviour is more understandable, than array of errors.

juljaVysotska (changes requested):
Good job! I found some moments, please check and fix it (especially add file convertToCase.js).
Best regards 💛

src/errorMessages.js
vladyslav-mosor: I moved caseNames to a separate file, so I changed logic in errorMessages a little.

src/validateUrl.js
vladyslav-mosor: image
There is the requirement to red show errors like this to follow the payload, so I just think it may cause some tests problems.

src/createServer.js
vladyslav-mosor: Is it necessary to add since I didn`t make any changes to those files?

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+
+    const text = pathname.slice(1);
+    const toCase = searchParams.get('toCase');
+
+    const sendResponse = (code, message, data) => {
+      res.statusCode = code;
+      res.statusMessage = message;
+      res.setHeader('Content-Type', 'application/json');
+      res.end(JSON.stringify(data));
+    };
```
OleksiiBlezniuk, line 18: Good idea, I think if you'll create a factory of such helpers it will be more useful on big projects

OleksiiBlezniuk (approved):

PR #235, opened 2023-05-02
solution

milenadiamonds1:


valeriiayarzhemska (approved):

src/checkUrl.js
```diff
@@ -0,0 +1,34 @@
+const cases = [
+  'SNAKE',
+  'KEBAB',
+  'CAMEL',
+  'PASCAL',
+  'UPPER',
+];
```
OleksiiBlezniuk, line 7: Usually, we move such constants to separate files and use uppercased names for them

src/createServer.js
```diff
@@ -1,3 +1,53 @@
...
+        JSON.stringify({
+          errors,
+        }),
+      );
+
+      return;
+    }
+
+    res.statusCode = 200;
+    res.statusMessage = 'OK';
```
OleksiiBlezniuk, line 33: it's a default status code, so you can skip it

PR #236, opened 2023-05-02
done

kateryna-babina:


src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+      res.statusCode = 400;
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const newCase = convertToCase(text, toCase);
+
+    res.end(JSON.stringify(
+      {
```
Ambulence, line 26: Respond to the client
You should respond with status 200 and OK status text.

Ambulence (approved):
Great!

src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+
+    const toCase = normalizedURL.searchParams.get('toCase');
+    const text = normalizedURL.pathname.slice(1);
+
+    res.setHeader('Content-Type', 'application/json');
+
+    const errors = handleErrors(text, toCase);
```
OleksiiBlezniuk, line 14: Do you throw any errors to handle here? Use more descriptive names, this function should **validate params**, so better to call it this way

src/createServer.js
OleksiiBlezniuk: 200 OK - is a default status code in the `response` object

src/handleErrors.js
```diff
@@ -0,0 +1,22 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const correctRequest = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+
+function handleErrors(text, toCase) {
```
OleksiiBlezniuk, line 4:
```suggestion
function validateParams(text, toCase) {
```

OleksiiBlezniuk (approved):
Will approve, but pay attention to naming

PR #238, opened 2023-05-02
Solution

DenysDubas:


src/checkUrlParams.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable max-len */
+const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const checkUrlParams = (text, toCase) => {
+  const errors = [];
+
+  if (!text) {
+    errors.push({
+      message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
OleksiiBlezniuk, line 9: Will be more useful to move such text to constant to avoid magic string and to maintain it in one place

src/createServer.js
```diff
@@ -1,3 +1,47 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/');
+const { checkUrlParams } = require('./checkUrlParams');
+
+const getParamsFromUrl = (url) => {
```
OleksiiBlezniuk, line 5: You can move it to separate file

OleksiiBlezniuk (changes requested):

PR #239, opened 2023-05-02
Add createServer function for case conversion server

gretoq:


src/createServer.js
```diff
@@ -1,3 +1,43 @@
...
+
+      return;
+    }
+
+    const {
+      originalCase,
+      convertedText,
+    } = convertToCase(originalText, targetCase);
+
+    res.statusCode = 200;
```
OleksiiBlezniuk, line 32: It's a default status code, so you can skip it here

src/getErrors.js
```diff
@@ -0,0 +1,33 @@
+const CASES = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
OleksiiBlezniuk, line 1: Usually, we move such constants to separate file

src/getErrors.js
```diff
@@ -0,0 +1,33 @@
...
+
+const example = '/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>';
+
+const messageTextDoesNotExist = `Text to convert is required. Correct request is: "${example}".`;
+const messageCaseDoesNotExist = `"toCase" query param is required. Correct request is: "${example}".`;
+const messageCaseTypeIsNotValid = `This case is not supported. Available cases: ${CASES.join(', ')}.`;
```
OleksiiBlezniuk, line 7: These as well

OleksiiBlezniuk (approved):

PR #240, opened 2023-05-02
CaseTransformerServer task is done

MarkBabich:


src/getValidationErrors.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable max-len */
+const cases = [
+  'SNAKE',
+  'KEBAB',
+  'CAMEL',
+  'PASCAL',
+  'UPPER',
+];
```
OleksiiBlezniuk, line 8: Usually, we move such constants to separate file and use uppercased name

src/getValidationErrors.js
```diff
@@ -0,0 +1,34 @@
...
+
+function getValidationErrors(textToConvert, toCase) {
```
OleksiiBlezniuk, line 10: I think purpose of this function is to validate data, not to get errors

OleksiiBlezniuk (approved):

PR #241, opened 2023-05-02
:white_check_mark: implement case transformer

Bogdan-Mykhailov:
:white_check_mark: implement case transformer

### DIFF
--- a/src/convertToCase/validateCaseErrors.js
+++ b/src/convertToCase/validateCaseErrors.js
@@ -1,0 +1,22 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const correctRequest = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+
+function validateCaseErrors(text, toCase) {
+  const errors = [];
+
+  if (!text) {
+    errors.push({ message: `Text to convert is required. Correct request is: ${correctRequest}` });
+  }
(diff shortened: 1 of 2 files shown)


src/convertToCase/validateCaseErrors.js
```diff
@@ -0,0 +1,22 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const correctRequest = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
```
OleksiiBlezniuk, line 2: Usually, we move constants to separate file and use uppercased names for them

src/convertToCase/validateCaseErrors.js
```diff
@@ -0,0 +1,22 @@
...
+
+function validateCaseErrors(text, toCase) {
+  const errors = [];
+
+  if (!text) {
+    errors.push({ message: `Text to convert is required. Correct request is: ${correctRequest}` });
+  }
```
OleksiiBlezniuk, line 9:
```suggestion
  if (!text) {
    errors.push({
      message: `Text to convert is required. Correct request is: ${correctRequest}`,
    });
  }
```

OleksiiBlezniuk: Line looks too long

OleksiiBlezniuk (approved):

PR #242, opened 2023-05-02
Solution v1

yaroslav-vasilyev:


Stepan-TS (approved):

src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+
+  const errors = checkErrors(textToConvert, toCase);
+
+  if (errors.length) {
+    res.statusCode = 400;
+
+    res.end(JSON.stringify({
+      errors,
+    }));
+  } else {
```
OleksiiBlezniuk, line 20: you can return from function in if block and remove this else block

OleksiiBlezniuk (approved):

PR #243, opened 2023-05-02
Develop node_case-transformer-server

sergey-lipikhin:


andriibezkorovainyi (approved):
GJ!

src/responses.js
```diff
@@ -0,0 +1,24 @@
+function respondWithError(res, errors) {
+  res.statusCode = 400;
+  res.statusMessage = 'Bad request';
```
OleksiiBlezniuk, line 3:
```suggestion
function respondWithBadRequestError(res, errors) {
  res.statusCode = 400;
  res.statusMessage = 'Bad request';
```

OleksiiBlezniuk: errors can be different so for better reusability better to either create helper for a particular error or pass the error code as parameter

OleksiiBlezniuk (approved):

PR #246, opened 2023-05-03
node_case-transormer-server

Stepan-TS:


src/validation.js
```diff
@@ -0,0 +1,30 @@
...
+    errors.push({
+      message: 'This case is not supported. '
+        + `Available cases: ${cases.join(', ')}.`,
+    });
+  }
+
+  return errors;
+};
+
+module.exports = { isValideUrl };
```
juljaVysotska, line 30: Typo in exported function name 'isValideUrl'. Rename it to 'isValidUrl'.

src/validation.js
```diff
@@ -0,0 +1,30 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const isValideUrl = (text, toCase) => {
```
juljaVysotska, line 3: Typo in function name 'isValideUrl'. Rename it to 'isValidUrl'.

src/createServer.js
```diff
@@ -1,3 +1,32 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const { convertToCase } = require('./convertToCase/convertToCase.js');
+const http = require('http');
+const { isValideUrl } = require('./validation.js');
```
juljaVysotska, line 3: Typo in the function name 'isValideUrl', it should be 'isValidUrl'.

src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+
+const createServer = () => {
+  return http.createServer((req, res) => {
+    const normaizeUrl = new URL(req.url, `http://${req.headers.host}`);
```
juljaVysotska, line 7: Instead of 'normaizeUrl', use 'normalizeUrl' for better readability.

src/validation.js
```diff
@@ -0,0 +1,30 @@
...
+  const errors = [];
+
+  if (!text) {
+    errors.push({
+      message: 'Text to convert is required. '
```
juljaVysotska, line 8: It's a bad way to represent string. If you have trouble with linter, you can add eslint-disable rule or change eslint rule.

src/validation.js
```diff
@@ -0,0 +1,30 @@
...
+  };
+
+  if (!cases.includes(toCase) && toCase) {
+    errors.push({
+      message: 'This case is not supported. '
+        + `Available cases: ${cases.join(', ')}.`,
+    });
+  }
+
+  return errors;
```
juljaVysotska, line 27: If this function validates data, maybe makes sense to throw a new error if validation failed? This function can return nothing, BUT throw error if array of errors is not empty. This behaviour is more understandable, than an array of errors.

src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+      res.statusCode = 400;
+      res.statusText = 'Bad request';
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const convertedText = convertToCase(originalText, targetCase);
+
+    res.end(JSON.stringify({
```
juljaVysotska, line 24: Good practice will be set statusCode and statusText too.

juljaVysotska (changes requested):
Good job! I like your code! I found some issues, please check and fix them. Best regards 💛

PR #247, opened 2023-05-03
solution

Nordexz:


src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+
+const createServer = () => {
+  return http.createServer((req, res) => {
+    const normilizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = normilizedURL.pathname.slice(1);
+    const targetCase = normilizedURL.searchParams.get('toCase');
+
+    res.setHeader('Content-Type', 'application/json');
+
+    const errors = validateError(originalText, targetCase);
```
Oleksandr-kopaevich, line 13: You don't validate errors, but incoming arguments. The part of the name Error is incorrect here.

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+
+    if (errors.length !== 0) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const convertedText = convertToCase(originalText, targetCase);
```
Oleksandr-kopaevich, line 23: When I see a text in the naming I expect that it's just a string, but then you spread the string into a new object... If it's an object then don't name it like text.

src/validateError.js
```diff
@@ -0,0 +1,39 @@
...
+const errorMessage = {
+  textIsMissing: {
+    message: 'Text to convert is required. ' + errorTextEndpoint,
+  },
+  caseIsMissing: {
+    message: '"toCase" query param is required. ' + errorTextEndpoint,
+  },
+  unknownCase: {
+    message: 'This case is not supported. '
+    + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
Oleksandr-kopaevich, line 13: Use here values from the variable `cases` and `join` method.

Oleksandr-kopaevich (approved):
Hey @Nordexz, good job, all tests have passed. The code is approved, but please be more mindful about naming variables/functions.

PR #248, opened 2023-05-03
case-transformer-server

Ambulence:


src/validation.js
```diff
@@ -0,0 +1,22 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const urlFormat = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>"';
```
And678, line 2: Nice! Keeping it DRY :fire:

src/createServer.js
```diff
@@ -1,3 +1,34 @@
...
+      return;
+    }
+
+    res.writeHead(200, 'OK', { 'Content-Type': 'application/json' });
+
+    const {
+      originalCase, convertedText,
+    } = convertToCase(textToConvert, toCase);
+
+    res.end(JSON.stringify({
```
And678, line 25: `.writeHead` returns `res`, so you can chain it like this:

```suggestion
    const {
      originalCase, convertedText,
    } = convertToCase(textToConvert, toCase);

    res
      .writeHead(200, 'OK', { 'Content-Type': 'application/json' })
      .end(JSON.stringify({
```

And678 (approved):
Good job, clean and simple. Approve :heavy_check_mark:! Check one suggestion below)

PR #249, opened 2023-05-03
node case transformer server

andriibezkorovainyi:


src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+  }
+
+  const response = {
+    ...convertToCase(textToConvert, toCase),
+    targetCase: toCase,
+    originalText: textToConvert,
+  };
+
+  res.statusCode = 200;
+  res.statusText = 'OK';
```
OleksiiBlezniuk, line 54: these are default status code and text, you can skip it

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+  const toCase = normalizedUrl.searchParams.get('toCase');
+
+  const errorMessages = {
+    errors: [],
+  };
+
+  if (!textToConvert) {
+    errorMessages.errors.push({
+      // eslint-disable-next-line max-len
+      message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
OleksiiBlezniuk, line 19: I think better to move this correct request to constant

OleksiiBlezniuk (changes requested):

PR #250, opened 2023-05-03
node case transformer - solution

Yevhenii0536:


src/CheckerErrors.js
And678, line 1: File name should be the [same as its main export](https://github.com/mate-academy/style-guides/blob/master/javascript.md#237use-camelcase-when-you-export-a-function-your-filename-should-be-identical-to-your-functions-name). This file should be called `checkerErrors.js` (lowercase letter).

src/CheckerErrors.js
```diff
@@ -0,0 +1,30 @@
...
+    errors.push({
+      // eslint-disable-next-line max-len
+      message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+    });
+  }
+
+  if (!messageToConvert) {
+    errors.push({
+      // eslint-disable-next-line max-len
+      message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
And678, line 16: Save your correct format (`"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>"`) to a const variable, and put it in message using a [template string literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals). 

That way, you will not repeat it, and you will not need to make eslint disable comments.
For the `This case is not supported...` message, you can put `convertParameters.join(', ')` inside a template string.

And678 (changes requested):
Good job, almost done! But some style changes are needed.

PR #251, opened 2023-05-03
solution

abcdesign1:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+  return http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const { url } = req;
+    const urlSplit = url.split('?');
+    const params = new URLSearchParams(urlSplit[1]);
+    const textToConvert = urlSplit[0].slice(1);
+    const toCase = params.get('toCase');
+
+    const caseNames = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
OleksiiBlezniuk, line 16: Better to move it outside this function, it's a static constant, so best option is to move it to separate file (at least outside function)

src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+        errorsArr.push(ErrorMessages.noCaseParam);
+      }
+
+      if (!caseNames.includes(toCase) && toCase) {
+        errorsArr.push(ErrorMessages.notValidCase);
+      }
+
+      if (errorsArr.length !== 0) {
+        throw new Error(JSON.stringify(errorsArr));
+      }
```
OleksiiBlezniuk, line 35: Better to move this validation to separate function

src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+
+      const responseToSend = {
+        originalCase: convertedText.originalCase,
+        targetCase: toCase,
+        originalText: textToConvert,
+        convertedText: convertedText.convertedText,
+      };
+
+      res.statusCode = 200;
+      res.statusMessage = 'OK';
```
OleksiiBlezniuk, line 47: these are default status code and message, you can skip it

src/utils/ErrorMessages.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable max-len */
+const ErrorMessages = {
```
OleksiiBlezniuk, line 2:
```suggestion
const ERROR_MESSAGES = {
```

OleksiiBlezniuk: Usually we use uppercased snake case for such constants names

src/utils/ErrorMessages.js
```diff
@@ -0,0 +1,16 @@
...
+  noTextToConvert: {
+    message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  },
+  noCaseParam: {
+    message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  },
+  notValidCase: {
+    message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
OleksiiBlezniuk, line 10: What if available cases will be changed? Use values from cases array instead

src/utils/ErrorMessages.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable max-len */
+const ErrorMessages = {
+  noTextToConvert: {
+    message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
OleksiiBlezniuk, line 4: I guess will be useful to move this correct request to constant

src/httpGet.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable no-console */
+const http = require('http');
+
+const PORT = process.env.PORT || 5700;
+
+http.get(`http://localhost:${PORT}/fdgh-dfsg?toCase=KEBA`, (res) => {
```
OleksiiBlezniuk, line 6: Why do you need this file?

OleksiiBlezniuk (changes requested):

PR #253, opened 2023-05-03

stanislavkulakovskyi:


src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { handleErrors } = require('./handleErrors');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const originalText = req.url.split('?').slice(0, 1).join('').slice(1);
+    const queryString = req.url.split('?').slice(1).join('');
```
juljaVysotska, line 8: Remove this line, as you already have queryString from line 10.

src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
+    const params = new URLSearchParams(queryString);
+    const targetCase = params.get('toCase');
```
juljaVysotska, line 10: Use the 'has' method to check if 'toCase' parameter exists before getting its value.

src/createServer.js
```diff
@@ -1,3 +1,36 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { handleErrors } = require('./handleErrors');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const originalText = req.url.split('?').slice(0, 1).join('').slice(1);
```
juljaVysotska, line 7: Use destructuring assignment to get the originalText and queryString from req.url.

src/createServer.js
```diff
@@ -1,3 +1,36 @@
...
+    const targetCase = params.get('toCase');
+    const errors = handleErrors(originalText, targetCase);
+
+    if (errors.length) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+      res.setHeader('Content-Type', 'application/json');
+      res.end(JSON.stringify({ errors }));
+    } else {
+      res.statusCode = 200;
```
juljaVysotska, line 19: You have duplicated code here and on 14-17 lines. We can create a function with arguments to avoid duplication. That function just receives status code, message and data for JSON.stringify into arguments. It will be easily to use and reuse this function in future.

src/handleErrors.js
```diff
@@ -0,0 +1,34 @@
+const handleErrors = (text, caseName) => {
+  const caseNames = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+  const errors = [];
+
+  if (!text) {
+    errors.push({
+      message:
+        'Text to convert is required. Correct request is: '
```
juljaVysotska, line 8: It's a bad way to represent string. If you have trouble with linter, you can add eslint-disable rule or change eslint rule.

juljaVysotska (changes requested):
Good job! I found some moments, please check and fix it.
Best regards 💛

PR #255, opened 2023-05-03
node case transform solution

ksavelyeva:


src/createServer.js
```diff
@@ -1,3 +1,44 @@
...
+      errors.push({ message: '"toCase" query param is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
+    } else if (!SUPPORTED_CASES.includes(targetCase)) {
+      errors.push({ message: 'This case is not supported. '
+        + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.' });
+    }
+
+    if (errors.length) {
+      res.statusCode = 400;
+      res.setHeader('Content-Type', 'application/json');
```
And678, line 29: You can set this header at the top of this function, as you need to always return json

And678 (approved):
Approve :heavy_check_mark: , but check out my comments about style

PR #257, opened 2023-05-04
done task

BozhenaBodnar:


src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+    res.writeHead(200, { 'Content-Type': 'application/json' });
+
+    res.end(
+      JSON.stringify({
+        originalCase: result.originalCase,
+        targetCase: toCase,
+        originalText: text,
+        convertedText: result.convertedText,
+      }),
+    );
```
And678, line 60: Can be simplified a little, [writeHead](https://nodejs.org/api/http.html#responsewriteheadstatuscode-statusmessage-headers) already return `res` so you can chain it.
```suggestion
    res
      .writeHead(200, { 'Content-Type': 'application/json' })
      .end(
        JSON.stringify({
          ...result,
          targetCase: toCase,
          originalText: text,
        }),
    );
```

src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+      errors.push({
+        message: '"toCase" query param is required. Correct request is: '
+        + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    if (!supportedCases.includes(toCase) && toCase) {
+      res.writeHead(400, { 'Content-Type': 'application/json' });
```
And678, line 31: No need to call this here, because you do it before your `res.end`
```suggestion
```

src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+    if (!toCase) {
+      errors.push({
+        message: '"toCase" query param is required. Correct request is: '
+        + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    if (!supportedCases.includes(toCase) && toCase) {
```
And678, line 30: Better move your data validation logic to a separate module, code will look cleaner :)

And678 (changes requested):
Good, only a couple of things to fix and this is done :100:

PR #259, opened 2023-05-04
Node Case Tranformer Server

VadKol:


src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+
+const { getUrlParameters } = require('./URLParameters');
+const { checkURL } = require('./catchErrors');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const [text, query] = req.url.split('?');
```
juljaVysotska, line 11: Use the 'url' module to properly parse the URL and query parameters.

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+      res.end(
+        JSON.stringify({
+          errors,
+        }),
+      );
+
+      return;
+    }
+
+    res.statusCode = 200;
```
juljaVysotska, line 28: You have duplicated code here and on 16-19 lines. We can create a function with arguments to avoid duplication. That function just receives status code, message and data for JSON.stringify into arguments. It will be easily to use and reuse this function in future.

src/catchErrors.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+  }
+
+  if (!cases.includes(toCase) && toCase) {
+    errors.push({
+      message: `This case is not supported. Available cases: ${cases.join(', ')}.`,
+    });
+  }
+
+  return errors;
```
juljaVysotska, line 25: If this function validates data, maybe makes sense to throw a new error if validation failed? This function can return nothing, BUT throw error if array of errors is not empty. This behaviour is more understandable, than array of errors.

juljaVysotska (changes requested):
Good job! I found some issues, please check and fix it. Best regards 💛

src/createServer.js
VadKol: Fixed

src/createServer.js
VadKol: Fixed

Oleksandr-kopaevich (approved):
Hey @VadKol, tests are passed and all main comments are fixed, approved

PR #264, opened 2023-05-30
solved

PashaMalyshkin:


src/createServer.js
```diff
@@ -1,3 +1,53 @@
...
+      }
+    }
+
+    if (!validation.errors.length) {
+      res.end(JSON.stringify(result));
+    } else {
+      res.statusCode = 400;
+      res.end(JSON.stringify(validation));
+    }
+  });
```
SerhiiKirik, line 49: Create a separate function to send response
```suggestion
      if (validation.errors.length) {
        sendResponse(400, result);
        
        return;
     }
     
      const result = {
        ...convertToCase(queryString, stringCase),
        originalText: queryString,
        targetCase: stringCase,
      }
      
     sendResponse(200, result);
  });
```

SerhiiKirik (changes requested):
Well done 🔥 

src/createServer.js
```diff
@@ -1,3 +1,33 @@
...
+        originalText: queryString,
+        targetCase: stringCase,
+      };
+
+      sendResponse(res, 200, result);
+
+      return;
+    }
+
+    if (validation.errors.length) {
```
SerhiiKirik, line 26: you already check it in 14 like
if thick check === false `If (!validation.errors.length)`  it means that `validation.errors.length` === true )
```suggestion
```

SerhiiKirik (approved):
Well done 🔥

PR #267, opened 2023-05-30
created server

Sbavdik:


src/sendResponse.js
```diff
@@ -0,0 +1,11 @@
+const sendResponse = (res, response, statusCode, data) => {
+  res.setHeader('Content-Type', 'application/json');
+  res.statusCode = statusCode;
+  res.statusText = response;
+
+  res.end(data);
```
SerhiiKirik, line 6: JSON.stringify should be used to convert data to a JSON string before sending it in the response.

src/standartizeUrl.js
```diff
@@ -0,0 +1,11 @@
+const standartizeURL = (url) => {
+  const textToConvert = url.split('?');
```
SerhiiKirik, line 2: This function is responsible for parsing and extracting parameters from the URL, but it is not converting the text between cases as per the task description. Consider adding the conversion logic or calling another function that handles the conversion.

main.js
```diff
@@ -5,3 +5,5 @@ createServer()
     // eslint-disable-next-line no-console
     console.log('Server started! 🚀');
   });
+
+module.exports = { createServer };
```
SerhiiKirik, line 9: why do you need to export it here?
```suggestion
```

src/checkErrors.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable max-len */
+const errors = [
```
SerhiiKirik, line 2:
```suggestion
const errorMessages = [
```

src/checkErrors.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!url.length) {
+    errorMessage.push(errors[0]);
+  }
+
+  if (!toCase) {
+    errorMessage.push(errors[1]);
+  }
+
+  if (!['SNAKE', 'KEBAB', 'PASCAL', 'CAMEL', 'UPPER'].includes(toCase) && toCase) {
```
SerhiiKirik, line 25: Move it to the variable
```suggestion
  if (!....includes(toCase) && toCase) {
```

src/standartizeUrl.js
```diff
@@ -0,0 +1,11 @@
+const standartizeURL = (url) => {
```
SerhiiKirik, line 1:
```suggestion
const standardizeURL = (url) => {
```

src/standartizeUrl.js
```diff
@@ -0,0 +1,11 @@
...
+  const textToConvert = url.split('?');
+  const params = new URLSearchParams(textToConvert[1]);
+  const toCase = params.get('toCase');
+
+  return [textToConvert[0].slice(1), toCase];
```
SerhiiKirik, line 6: Make it like a object, it would be better
```suggestion
  return { ...: textToConvert[0].slice(1), ...: toCase };
```

src/createServer.js
```diff
@@ -1,3 +1,38 @@
...
+    sendResponse(
+      res,
+      'OK',
+      200,
+      JSON.stringify({
+        originalCase: convertedText.originalCase,
+        targetCase: paramsToConvert[1],
+        originalText: paramsToConvert[0],
+        convertedText: convertedText.convertedText,
+      }),
```
SerhiiKirik, line 29:
```suggestion
      {
        originalCase: convertedText.originalCase,
        targetCase: paramsToConvert[1],
        originalText: paramsToConvert[0],
        convertedText: convertedText.convertedText,
      },
```

SerhiiKirik (changes requested):
Well done 🚀 

src/standartizeUrl.js
Sbavdik: i tried to do this, but when i used convertToCase function in standardizeURL its broke tests.

src/checkErrors.js
```diff
@@ -0,0 +1,35 @@
...
+  {
+    message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  },
+  {
+    message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  },
+  {
+    message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+  },
+];
```
SerhiiKirik, line 12: Make it with object
```suggestion
const errorMessages = [
    SomeError: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
    SomeError: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
    SomeError: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
};
```

src/standartizeUrl.js
Sbavdik: i can't do this because convertToCase function accept 2 arguments instead of object

SerhiiKirik (approved):
LGTM 🔥

PR #268, opened 2023-05-30
create server

VenomMB:


src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+
+const http = require('http');
+const { checkOnError } = require('./checkOnError');
+const { createResponseBody } = require('./createResponseBody');
+const { toSplitUrl } = require('./splitUrl');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const queryString = req.url.split('?');
+    const [text, targetCase] = toSplitUrl(queryString);
```
SerhiiKirik, line 10: It's better to use destructuring assignment with the 'toSplitUrl' function result, rather than using array indexes.

src/splitUrl.js
```diff
@@ -0,0 +1,13 @@
+function toSplitUrl(query) {
+  const text = query[0].slice(1);
```
SerhiiKirik, line 2: Instead of using array indexing to access query parameters, use destructuring assignment. You can change it to: `const [text, targetCase] = query;` and then remove lines 2 and 3.

src/checkOnError.js
```diff
@@ -0,0 +1,32 @@
+/* eslint-disable max-len */
+function checkOnError(text, targetCase) {
+  const errorMessages = [
+    'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+  ];
```
SerhiiKirik, line 7: Would be better to do it with an object
For example:
But change keys!
```suggestion
  const errorMessages = {
    Text: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
    TargetCase: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
    NotSupported: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
  };
```

src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+    const errorBody = checkOnError(text, targetCase);
+    const hasError = errorBody.errors.length !== 0;
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = hasError ? 400 : 200;
+    res.statusMessage = hasError ? 'Bad request' : 'OK';
+
+    if (hasError) {
+      res.end(JSON.stringify(errorBody));
+    } else {
```
SerhiiKirik, line 20: try to avoie `if / else` construction?

src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const queryString = req.url.split('?');
+    const [text, targetCase] = toSplitUrl(queryString);
+    const errorBody = checkOnError(text, targetCase);
+    const hasError = errorBody.errors.length !== 0;
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = hasError ? 400 : 200;
```
SerhiiKirik, line 15: would be better to do it just with one `if`

SerhiiKirik (changes requested):
Great 🔥 

src/createServer.js
VenomMB: I didn't get it.

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+    const errorBody = checkOnError(text, targetCase);
+    const hasError = errorBody.errors.length !== 0;
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = 200;
+    res.statusMessage = 'OK';
+
+    if (hasError) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
```
SerhiiKirik, line 20: DRY
you can create separate function for the `sendResponse` & reuse it

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+
+      return res.end(JSON.stringify(errorBody));
+    }
+
+    const responseBody = createResponseBody(text, targetCase);
```
SerhiiKirik, line 25:
```suggestion
    if (hasError) {
      res.statusCode = 400;
      res.statusMessage = 'Bad request';

      return res.end(JSON.stringify(errorBody));
    }
    
    res.statusCode = 200;
    res.statusMessage = 'OK';

    const responseBody = createResponseBody(text, targetCase);
```

SerhiiKirik (approved):
Great 🔥

PR #269, opened 2023-05-30
solution

Olekk17:


src/createServer.js
```diff
@@ -1,3 +1,50 @@
...
+
+    res.setHeader('Content-Type', 'application/json');
+
+    if (!text) {
+      responseErrors.errors.push({ message: errorMessages.missingText });
+    }
+
+    if (!targetCase) {
+      responseErrors.errors.push({ message: errorMessages.missingTargetCase });
+    }
```
SerhiiKirik, line 23: Move it to the separate function
```suggestion
    res.setHeader('Content-Type', 'application/json');
```

src/createServer.js
```diff
@@ -1,3 +1,50 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const errorMessages = require('./errorMessages');
```
SerhiiKirik, line 3: you can make a destructuring from this object
```suggestion
const { 
  missingText, 
  missingTargetCase, 
  wrongTargetCase 
} = require('./errorMessages');
```

src/createServer.js
```diff
@@ -1,3 +1,41 @@
...
+
+    try {
+      Object.assign(response, convertToCase(text, targetCase));
+    } catch {
+      if (targetCase) {
+        responseErrors.errors.push({ message: wrongTargetCase });
+      }
+    }
+
+    if (!responseErrors.errors.length) {
```
SerhiiKirik, line 27: Would be better to another logic
First, you try to handle the error - then you make a succusses logic

SerhiiKirik (approved):
Well done 🔥

PR #272, opened 2023-05-30
node case transformer server is solved

M1k1ta:


src/createServer.js
```diff
@@ -1,3 +1,57 @@
...
+
+    res.setHeader('Content-Type', 'application/json');
+
+    if (!text) {
+      responseErrors.errors.push({ message: errorMessages.textError });
+    }
+
+    if (!toCase) {
+      responseErrors.errors.push({ message: errorMessages.caseError });
+    }
```
SerhiiKirik, line 27: Let's Decompose your logic
Move this logic to the separate function
```suggestion
    const ... = ...(...);
```

src/createServer.js
```diff
@@ -1,3 +1,57 @@
...
+
+    const respond = {
```
SerhiiKirik, line 29:
```suggestion
    const response = {
```

src/createServer.js
```diff
@@ -1,3 +1,57 @@
...
+    } catch {
+      if (toCase) {
+        responseErrors.errors.push({ message: errorMessages.caseInvalid });
+      }
+    }
+
+    if (!responseErrors.errors.length) {
+      res.statusCode = 200;
+      res.end(JSON.stringify(respond));
+    } else {
```
SerhiiKirik, line 45: try to avoid `if / else` construction
Can you do it without this construction?

SerhiiKirik (changes requested):
Great 🔥 
Let's decompose your logic 🧑‍💻

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your solution 🧑‍💻

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+    const response = {
+      originalText: text,
+      targetCase: toCase,
+    };
+
+    const responseError = findError(response);
+
+    if (responseError.errors.length) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
```
SerhiiKirik, line 20: Also, you can create separate function to send response & reuse it

SerhiiKirik (approved):
LGTM 🔥

PR #274, opened 2023-05-30
Create case transformer server

AnyaLyaLya:


src/constants/errorsMessages.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable max-len */
+const errorMessages = [
```
SerhiiKirik, line 2: would be better to do it with object
Try do it

src/createResponse.js
```diff
@@ -0,0 +1,14 @@
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+function createResponse(text, toCase) {
+  const convertResult = convertToCase(text, toCase);
+
+  return {
```
SerhiiKirik, line 6: wrap it with `JSON. stringify`

src/parseUrlQuery.js
```diff
@@ -0,0 +1,8 @@
+function parseUrlQuery(urlQuery) {
+  const [text, targetCase] = [urlQuery[0].slice(1), urlQuery[1]];
+  const toCase = new URLSearchParams(targetCase).get('toCase');
+
+  return [text, toCase];
```
SerhiiKirik, line 5: would be better to return an object here

src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+    const errors = errorCheck(text, toCase);
+    const isError = errors.length !== 0;
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = isError ? 400 : 200;
+    res.statusMessage = isError ? 'Bad request' : 'OK';
+
+    if (isError) {
+      res.end(JSON.stringify({
+        errors: errors.map((error) => ({ message: error })),
```
SerhiiKirik, line 21: instead to do it, you can just push `{ message: ... }` in the `errorCheck` function & you avoid extra ineration

src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+    const urlQuery = req.url.split('?');
+
+    const [text, toCase] = parseUrlQuery(urlQuery);
+
+    const errors = errorCheck(text, toCase);
+    const isError = errors.length !== 0;
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = isError ? 400 : 200;
+    res.statusMessage = isError ? 'Bad request' : 'OK';
```
SerhiiKirik, line 17: would be better to do it with 1 `if`

src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+
+    if (isError) {
+      res.end(JSON.stringify({
+        errors: errors.map((error) => ({ message: error })),
+      }));
+    } else {
```
SerhiiKirik, line 23: can you do it without `if / else` construction?

SerhiiKirik (changes requested):
Great 🔥 

src/createServer.js
```diff
@@ -1,3 +1,28 @@
...
+
+    res.setHeader('Content-Type', 'application/json');
+
+    res.statusCode = isError ? 400 : 200;
+    res.statusMessage = isError ? 'Bad request' : 'OK';
+
+    res.end(isError
+      ? JSON.stringify(errorsMessages)
+      : createResponse(text, toCase));
+  });
```
SerhiiKirik, line 23: You did this check 3 times)

```suggestion
    if (isError) {
       res.statusCode = 400;
       res.statusMessage = Bad request;
       res.end(JSON.stringify(errorsMessages));
       
       return;
    }
    
    res.statusCode = 200;
    res.statusMessage = 'OK';

    res.end(createResponse(text, toCase));
  });
```

SerhiiKirik (approved):
Nice 🔥

PR #277, opened 2023-05-30
create server solution

VladislavaT:


SerhiiKirik (changes requested):
Well done 🔥 

src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+const { sendRequest } = require('./utils/sendRequest');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const [textToFormat, caseToFormat] = normalizeUrl(req);
+    const errors = getErrors(textToFormat, caseToFormat);
+
+    if (errors.length > 0) {
+      sendRequest(res, 'Bad request', 400, { errors });
+    } else {
```
SerhiiKirik, line 14: try to avoid `if / else` construction, do it without this

src/utils/getErrors.js
```diff
@@ -0,0 +1,35 @@
...
+  },
+  {
+    message: `"toCase" query param is required. Correct request is: ${regs}.`,
+  },
+  {
+    message: `This case is not supported. Available cases: ${cases.join(
+      ', ',
+    )}.`,
+  },
+];
```
SerhiiKirik, line 15: make it like an object
It would be easy to take value from it
```suggestion
const errorMessages = {
    Error: `Text to convert is required. Correct request is: ${regs}.`,
    Error2: `"toCase" query param is required. Correct request is: ${regs}.`,
    Error3: `This case is not supported. Available cases: ${cases.join(', ',)}.`,
},;
```

src/createServer.js
```diff
@@ -1,3 +1,32 @@
...
+        textToFormat,
+        caseToFormat,
+      );
+
+      sendRequest(res, 'OK', 200, {
+        originalCase,
+        convertedText,
+        targetCase: caseToFormat,
+        originalText: textToFormat,
+      });
```
SerhiiKirik, line 25: move it to the variable for better readable
```suggestion
      sendRequest(res, 'OK', 200, ...);
```

SerhiiKirik (changes requested):
Keep it up 🔥 
Just a few improvements 👩‍💻

src/utils/getErrors.js
```diff
@@ -0,0 +1,27 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const regs = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>"';
+const errorMessages = {
+  Error: `Text to convert is required. Correct request is: ${regs}.`,
+  Error2: `"toCase" query param is required. Correct request is: ${regs}.`,
+  Error3: `This case is not supported. Available cases: ${cases.join(', ')}.`,
+};
```
SerhiiKirik, line 7: It was example 😄 
```suggestion
  Text: `Text to convert is required. Correct request is: ${regs}.`,
  Case: `"toCase" query param is required. Correct request is: ${regs}.`,
  CaseSupported: `This case is not supported. Available cases: ${cases.join(', ')}.`,
};
```

SerhiiKirik (approved):
LGTM 🔥

src/utils/getErrors.js
VladislavaT: oh, ok 😄
I'll fix this

PR #292, opened 2023-06-01
Solution

DimaDvm:


src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+const { errorCatch } = require('./errorCatch');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const { text, toCase } = request(req);
+
+    const errors = errorCatch(text, toCase)
+
+    if (errors.length > 0) {
+      res.writeHead(400, { 'Content-Type': 'application/json' });
```
SerhiiKirik, line 13: Make a separate function for sending request & reuse it in all cases

src/err.js
```diff
@@ -0,0 +1,19 @@
...
+
+  if (!text) {
+    errors.push({ message: 'Text to convert is required' });
+  }
+
+  if (!toCase) {
+    errors.push({ message: '"toCase" query param is required' });
+  }
+
+  if (!['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'].includes(toCase)) {
```
SerhiiKirik, line 12: Move it to the variable
```suggestion
  if (!....includes(toCase)) {
```

src/err.js
```diff
@@ -0,0 +1,19 @@
+function errorCatch(text, toCase) {
+  const errors = [];
+
+  if (!text) {
+    errors.push({ message: 'Text to convert is required' });
```
SerhiiKirik, line 5: You can't pass the tests with this messages
Make a separate object with your all errors messages

SerhiiKirik (changes requested):
Well done 🚀 

PR #294, opened 2023-06-01
Create server ;(

Hamuud:


src/validation.js
```diff
@@ -0,0 +1,37 @@
...
+
+const postMessageError = (response, statusCode, errorMessage) => {
+  const errors = [];
+
+  errorMessage.forEach(error => {
+    errors.push(error);
+  });
+  response.statusCode = statusCode;
+  response.end(JSON.stringify({ errors }));
+}
```
SerhiiKirik, line 32: Why do you do that?
you just create a new array with the same objects)
```suggestion
const postMessageError = (response, statusCode, errorMessage) => {
  response.statusCode = statusCode;
  response.end(JSON.stringify({ errors }));
}
```

SerhiiKirik (changes requested):
Well done 🔥 

SerhiiKirik (approved):
LGTM 🔥

PR #295, opened 2023-06-01
finish :(

Rage1502:


src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { handleErrors } = require('./HandleErrors');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const [text, queryString] = req.url.split('?');
+    const params = new URLSearchParams(queryString);
+
+    const textToConvert = text.slice(1);
+    const toCase = params.get('toCase');
```
SerhiiKirik, line 11: Would be better to move it to separate function, let's do it
```suggestion
    const { textToConvert, toCase } = ...(...);
    
    OR
    
        const [textToConvert, toCase] = ...(...);
```

src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+    const errors = handleErrors(textToConvert, toCase);
+
+    if (errors.length) {
+      res.writeHead(400, 'Bad Request', { 'Content-Type': 'application/json' });
```
SerhiiKirik, line 15: DRY
make a separate function to sending a response & reuse it

SerhiiKirik (changes requested):
Great 🔥 

PR #297, opened 2023-06-01
solution

yapelenua:


src/createServer.js
```diff
@@ -1,3 +1,37 @@
...
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const { inputText, toCase } = requestData(req);
+
+    const errorMessages = getError(toCase, inputText);
+
+    if (errorMessages.errors.length > 0) {
+      res.statusCode = 400;
```
SerhiiKirik, line 15: DRY
Let's make a separate function for sending a response & reuse it

SerhiiKirik (changes requested):
Nice 🔥 
Let's improve your solution 🧑‍💻

PR #298, opened 2023-06-02
fix

asmd564:


src/createServer.js
```diff
@@ -1,3 +1,41 @@
...
+
+function createServer() {
+    const server = http.createServer((req, res) => {
+      const { text, toCase } = data(req);
+      const errors = validate(text, toCase);
+  
+      res.setHeader('Content-Type', 'application/json');
+  
+      if (errors.length) {
+        res.statusCode = 400;
```
SerhiiKirik, line 14: DRY
create a separate function for sending response & reuse it

SerhiiKirik (changes requested):
Well done 🔥 

PR #301, opened 2023-06-02
solution

artem-kyrylenko:


src/createServer.js
```diff
@@ -1,3 +1,34 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
```
SerhiiKirik, line 1: remove it
```suggestion
```

src/normalizeUrl.js
```diff
@@ -0,0 +1,9 @@
+function normalizeUrl(requestUrl, requestHost) {
```
SerhiiKirik, line 1: Why do you create separate files for one function?
Just create 1 file like a `helpers` & put to this file all function for the server

src/createServer.js
```diff
@@ -1,3 +1,34 @@
...
+
+const createServer = () => {
+  const server = http.createServer((request, response) => {
+    const { toCase, textToConvert } = normalizeUrl(request.url, request.headers.host);
+    const validation = validateParams(textToConvert, toCase);
+
+    if (validation.errors.length) {
+      response.writeHead(400, { 'Content-Type': 'application/json' });
+
+      return response.end(JSON.stringify(validation));
```
SerhiiKirik, line 16: DRY
Make a separate function for sending response

SerhiiKirik (approved):
Well done 🔥

PR #303, opened 2023-06-06
Solution

voronevska:


src/createResponse.js
```diff
@@ -0,0 +1,14 @@
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+function createResponse(text, toCase) {
```
liliia-livandovskaya, line 3: Consider using arrow function syntax for defining 'createResponse' function

src/errorCheck.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable max-len */
```
liliia-livandovskaya, line 1: Consider removing the `eslint-disable max-len` comment and ensure the lines do not exceed the maximum length specified in your ESLint configuration. This will improve code readability.

liliia-livandovskaya (approved):

PR #304, opened 2023-06-07
Solution

OleksandrOse:


src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
+'use strict';
+/* eslint-disable no-console */
+/* eslint-disable dot-notation */
+
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+const allowedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
DarkMistyRoom, line 8: move it to a separate file

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+    if (!toCase) {
+      validation.errors.push({
+        message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    if (!(allowedCases.includes(toCase)) && toCase) {
+      validation.errors.push({
+        message: `This case is not supported. Available cases: ${allowedCases.join(', ')}.`,
+      });
```
DarkMistyRoom, line 37: create an object with all errors messages and move it to a separate file

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+      targetCase: toCase,
+      originalText: textToTransform,
+      convertedText,
+    }));
+  });
+
+  return server;
+}
+
+createServer();
```
DarkMistyRoom, line 63: no need to call it here. just export

DarkMistyRoom (changes requested):

PR #305, opened 2023-06-09
solution

ivanm2706:


src/createServer.js
```diff
@@ -1,3 +1,25 @@
...
+    const search = req.url.split('?')[1].split('=');
+    const [searchName, searchParam] = search;
+
+    const { error, message } = isValidReq(text, searchName, searchParam);
+
+    const contentRes = !error
+      ? convertToCase(text, searchParam)
+      : { errors: [{ message }] };
+
+    res.end(JSON.stringify(contentRes));
```
mykhalenych, line 17: Set the 'Content-Type' header to 'application/json' to indicate that the response is a JSON object.

src/isValidRequest.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable max-len */
```
mykhalenych, line 1: delete it and use template string

src/isValidRequest.js
```diff
@@ -0,0 +1,31 @@
...
+
+function isValidReq(text, searchProperty, searchParam) {
```
mykhalenych, line 3: It's better to use more descriptive function name like 'isValidRequest' instead of 'isValidReq'.

src/isValidRequest.js
```diff
@@ -0,0 +1,31 @@
...
+  let error = false;
+  let message = '';
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
mykhalenych, line 6: Move the 'cases' array outside of the function to avoid recreating it every time the function is called.

mykhalenych (changes requested):

mykhalenych (approved):

PR #314, opened 2023-07-18
add node_case-transformer-server solution

DidenkoDenis306:


33Orange (approved):
Good job. 

A point to improve:
You can create an object with all your default error messages and use them as reference

PR #315, opened 2023-07-18
for review

Vasilkoff:


src/validation.js
```diff
@@ -0,0 +1,26 @@
+const validation = (res, text, toCase) => {
```
xiimiik, line 1: Function names should typically be verbs or verb phrases. Consider renaming the function from 'validation' to something like 'validateInput' to make its purpose clearer.

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+    const { pathname, searchParams } = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = pathname.slice(1);
+    const targetCase = searchParams.get('toCase');
+
+    res.setHeader('Content-Type', 'application/json');
+
+    const errors = validation(res, originalText, targetCase);
+
+    if (errors.length) {
+      res.statusCode = 400;
```
xiimiik, line 18: Consider sending a descriptive error message along with the 400 status code to provide more information about the error.

src/createServer.js
```diff
@@ -1,3 +1,39 @@
 // Write code here
 // Also, you can create additional files in the src folder
 // and import (require) them here
+const http = require('http');
+const { convertToCase } = require('../src/convertToCase/convertToCase');
+const { validation } = require('./validation');
```
xiimiik, line 6: Function 'validation' should ideally be a verb or verb phrase. Consider renaming it to 'validateInput' or similar.

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+const createServer = () => {
+  const server = http.createServer((req, res) => {
```
xiimiik, line 8: Return 404 status code if the path does not match the expected format. This can be done by adding a conditional to check the pathname before proceeding.

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+
+    if (errors.length) {
+      res.statusCode = 400;
+
+      res.end(JSON.stringify({
+        errors,
+      }));
+    } else {
+      const { originalCase, convertedText }
+        = convertToCase(originalText, targetCase);
```
xiimiik, line 25: It's unclear what 'originalCase' is used for, as it's not mentioned in the task description. Consider removing it if it's not required.

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+const http = require('http');
+const { convertToCase } = require('../src/convertToCase/convertToCase');
+const { validation } = require('./validation');
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const { pathname, searchParams } = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = pathname.slice(1);
+    const targetCase = searchParams.get('toCase');
+
+    res.setHeader('Content-Type', 'application/json');
```
xiimiik, line 13: Sending 'application/json' as Content-Type header before validation may result in unnecessary overhead in case of errors. Consider moving this line after validation, immediately before the response is sent.

xiimiik (approved):
Great job 🔥

PR #316, opened 2023-07-18
add case transformer server task solution

opiquem:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+      res.statusMessage = 'Bad request';
+
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const convertionResult = convertToCase(originalText, targetCase);
+
+    const { originalCase, convertedText } = convertionResult;
```
33Orange, line 27:
```suggestion
    const { originalCase, convertedText } = convertToCase(originalText, targetCase);

```

33Orange (approved):
Well done

PR #318, opened 2023-07-18
add task solution

KatyaZaremska:


src/sendResponse.js
```diff
@@ -0,0 +1,8 @@
+const sendResponse = (response, statusCode, data) => {
+  response.setHeader('Content-Type', 'application/json');
+  response.statusCode = statusCode;
+
```
33Orange, line 4: `You should respond with status 200 and OK status text.`
You've forgotten about status text

src/getData.js
```diff
@@ -0,0 +1,9 @@
+const getData = (request) => {
```
33Orange, line 1:
```suggestion
const parseRequest = (request) => {
```

src/validateData.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable max-len */
+function validateData(text, toCase) {
+  const validCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+  const errorMessages = {
+    emptyTextMessage: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    emptyCaseMessage: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    invalidCaseMessage: `This case is not supported. Available cases: ${validCases.join(', ')}.`,
+  };
```
33Orange, line 8: Create these variables outside the function to avoid re-creating whenever it is executed

src/createServer.js
```diff
@@ -1,3 +1,31 @@
...
+const { getData } = require('./getData');
+const { validateData } = require('./validateData');
+const { sendResponse } = require('./sendResponse');
+
+const createServer = () => {
+  const server = http.createServer((request, response) => {
+    const [text, toCase] = getData(request);
+    const errors = validateData(text, toCase);
+
+    if (errors.length > 0) {
```
33Orange, line 12:
```suggestion
    if (errors.length) {
```

33Orange (changes requested):
Good job.

A few moments to improve

PR #319, opened 2023-07-18
made task node-case-transformer-server

vladyslav-lapshyn:


src/createServer.js
```diff
@@ -1,3 +1,46 @@
...
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = normalizedURL.pathname.substring(1);
+    const queryParams = new URLSearchParams(normalizedURL.search);
+    const targetCase = queryParams.get('toCase');
+
+    const validationErrors = validateRequest(originalText, targetCase);
```
33Orange, line 14: You validate text here, not a request.
Use more descriptive name

33Orange (changes requested):
Almost done

A few moments to improve

src/createServer.js
```diff
@@ -1,3 +1,45 @@
...
+
+    if (paramsErrors.length) {
+      res.statusCode = 400;
+      res.write(JSON.stringify({ errors: paramsErrors }));
+      res.end();
+
+      return;
+    }
+
+    res.statusCode = 200;
```
33Orange, line 24:
```suggestion
```

src/createServer.js
```diff
@@ -1,3 +1,45 @@
...
+    const originalText = normalizedURL.pathname.substring(1);
+    const queryParams = new URLSearchParams(normalizedURL.search);
+    const targetCase = queryParams.get('toCase');
+
+    const paramsErrors = validateParameters(originalText, targetCase);
+
+    if (paramsErrors.length) {
+      res.statusCode = 400;
+      res.write(JSON.stringify({ errors: paramsErrors }));
+      res.end();
```
33Orange, line 19: Same about status message and here you can provide JSON to the `end()` function too

33Orange (changes requested):
Almost done

33Orange (approved):
Well done

PR #320, opened 2023-07-19
solution

maryanmats:


src/validateUrlParams.js
```diff
@@ -0,0 +1,27 @@
+const validCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const isValidUrlParams = (text, toCase) => {
+  const errors = [];
+
+  if (!text) {
+    errors.push({
+      message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
33Orange, line 8: You can create an object with all your default error messages and use them as reference

33Orange (approved):
Well done

PR #325, opened 2023-07-19
add task solution

IrynaBalandiukh:


src/helpers.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable max-len */
+const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const isParamsValid = (textToConvert, targetCase) => {
+  const errors = [];
+
+  if (!textToConvert) {
+    errors.push({
+      message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
33Orange, line 9: You can create an object with all your default error messages and use them as reference

33Orange (approved):
Well done

PR #328, opened 2023-07-19
solution

vitvadym:


request.js
33Orange, line 1: What is this file? Should be here?

src/createServer.js
```diff
@@ -1,3 +1,108 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+const { errorMessage, caseName } = require('./errorMessage');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.writeHead(200, { 'Content-type': 'application/json' });
```
33Orange, line 7:
```suggestion
    res.setHeader('Content-type', 'application/json');
```

src/createServer.js
```diff
@@ -1,3 +1,108 @@
...
+      res.write(JSON.stringify({
+        errors: [
+          { message: errorMessage.caseNotSupported },
+        ],
+      }));
+
+      res.end();
+
+      return;
+    };
```
33Orange, line 87: Create a separate function that will take the `targetCase `and the `originalText` as arguments, validate them, and return the errors array.
After that check if you have any errors in the errors array and update the response
You have a lot of repeated logic.


You've forgotten about the status text.

src/createServer.js
```diff
@@ -1,3 +1,108 @@
...
+
+    res.statusCode = 200;
```
33Orange, line 89: `You should respond with status 200 and OK status text.`

src/createServer.js
```diff
@@ -1,3 +1,108 @@
...
+    const result = convertToCase(originalText, targetCase);
+
+    res.write(JSON.stringify({
+      originalCase: result.originalCase,
+      targetCase,
+      originalText,
+      convertedText: result.convertedText,
+    }));
+
+    res.end();
```
33Orange, line 100: You can combine these methods.

```suggestion
    res.end(JSON.stringify({
      originalCase,
      targetCase,
      originalText,
      convertedText,
    }));
```

src/createServer.js
```diff
@@ -1,3 +1,108 @@
...
+      }));
+
+      res.end();
+
+      return;
+    };
+
+    res.statusCode = 200;
+
+    const result = convertToCase(originalText, targetCase);
```
33Orange, line 91:
```suggestion
    const { originalCase, convertedText }= convertToCase(originalText, targetCase);
```

33Orange (changes requested):
Almost done

PR #329, opened 2023-07-19
add solution

growingananas:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+const { validateUrl } = require('./validateUrl');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const [textToConvert, toCase] = parseUrl(req);
+    const errors = validateUrl(textToConvert, toCase);
+
+    if (errors.length > 0) {
```
33Orange, line 15:
```suggestion
    if (errors.length) {
```

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+
+      const jsonResponse = {
+        originalCase,
+        targetCase: toCase,
+        originalText: textToConvert,
+        convertedText,
+      };
+
+      res.end(JSON.stringify(jsonResponse));
+    }
```
33Orange, line 34: Is it necessary to wrap this code in `else` construction?
You can use it as a default case.

```suggestion
    } 
    res.statusCode = 200;
    res.statusMessage = 'OK';

    const { originalCase, convertedText } = convertToCase(textToConvert, toCase);

    const jsonResponse = {
       originalCase,
       targetCase: toCase,
       originalText: textToConvert,
       convertedText,
    };

    res.end(JSON.stringify(jsonResponse));
```

33Orange (changes requested):
Good job.

A few moments with code style

33Orange (approved):
Well done

PR #330, opened 2023-07-19
solution

dimoshen:


src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+    const { pathname, searchParams } = normalizedURL;
+
+    const originalText = pathname.slice(1);
+    const targetCase = searchParams.get('toCase');
+
+    const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    const isSupportedCase = targetCase
+      ? supportedCases.includes(targetCase.toUpperCase())
+      : false;
```
33Orange, line 18:
```suggestion
```

src/setError.js
```diff
@@ -0,0 +1,32 @@
...
+    + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+
+    responseBody.errors.push({ message: errorMessage });
+
+    return responseBody;
+  }
+
+  if (!isSupportedCase) {
+    const errorMessage = 'This case is not supported. '
+    + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.';
```
33Orange, line 24: You can use a method join here

`supportedCases.join(', ')`

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+
+    if (!originalText || !targetCase || !isSupportedCase) {
```
33Orange, line 20: Is it neccesary?

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+      const responseError = setError(originalText, targetCase, isSupportedCase);
+
+      response.statusCode = 400;
+      response.setHeader('Content-Type', 'application/json');
+      response.end(JSON.stringify(responseError));
+
+      return;
```
33Orange, line 27: You can use the errors array to check if the errors exist.

```suggestion
      const errors = setError(originalText, targetCase);
      
      if (errors.length) {
        response.statusCode = 400;
        response.end(JSON.stringify(responseError));

        return;
      }
```

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+    const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    const isSupportedCase = targetCase
+      ? supportedCases.includes(targetCase.toUpperCase())
+      : false;
+
+    if (!originalText || !targetCase || !isSupportedCase) {
+      const responseError = setError(originalText, targetCase, isSupportedCase);
+
+      response.statusCode = 400;
```
33Orange, line 23: `If something is not correct, you should respond with 400 status, Bad request statusText `

Check the task requirements

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+      response.setHeader('Content-Type', 'application/json');
+      response.end(JSON.stringify(responseError));
+
+      return;
+    }
+
+    const convertionResults = convertToCase(originalText, targetCase);
+    const { originalCase, convertedText } = convertionResults;
+
+    response.statusCode = 200;
```
33Orange, line 33: `You should respond with status 200 and OK status text.`

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+    response.setHeader('Content-Type', 'application/json');
```
33Orange, line 34: You can set it in by default in both cases.
Right now you set it 2 times

src/createServer.js
```diff
@@ -1,3 +1,49 @@
...
+
+      response.statusCode = 400;
+      response.setHeader('Content-Type', 'application/json');
+      response.end(JSON.stringify(responseError));
+
+      return;
+    }
+
+    const convertionResults = convertToCase(originalText, targetCase);
+    const { originalCase, convertedText } = convertionResults;
```
33Orange, line 31:
```suggestion
    const { originalCase, convertedText } = convertToCase(originalText, targetCase);
```

33Orange (changes requested):
Almost done

PR #332, opened 2023-07-19
develop

kirilshorin:


src/validateUrl.js
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  if (!toCase) {
+    errors.push({
+      message: '"toCase" query param is required. Correct request is: '
+        + correctExample,
+    });
+  }
+
+  if (!cases.includes(toCase) && toCase) {
```
33Orange, line 21: Check is `toCase` exist first

```suggestion
  if (toCase && !cases.includes(toCase)) {
```

src/validateUrl.js
```diff
@@ -0,0 +1,32 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const validateURL = (text, toCase) => {
+  const errors = [];
+  const correctExample = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+
+  if (!text) {
+    errors.push({
+      message: 'Text to convert is required. Correct request is: '
```
33Orange, line 9: You can create an object with all your default error messages and use them as reference

src/validateUrl.js
```diff
@@ -0,0 +1,32 @@
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const validateURL = (text, toCase) => {
+  const errors = [];
+  const correctExample = '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
```
33Orange, line 5: You create this variable every time when the function is executed

src/createServer.js
```diff
@@ -1,3 +1,41 @@
...
+const http = require('http');
+const { validateURL } = require('./validateUrl');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const formatedPath = req.url.slice(1);
+    const [text, queryString] = formatedPath.split('?');
```
33Orange, line 10: You can use the `new URL` construction

```suggestion
    const formatedUrl = new URL(req.url, `http://${req.headers.host}`);
    const originalText = normalizedUrl.pathname.slice(1);
    const toCase = normalizedUrl.searchParams.get('toCase');
```

src/createServer.js
```diff
@@ -1,3 +1,41 @@
...
+
+    const params = new URLSearchParams(queryString);
+    const toCase = params.get('toCase');
+
+    const errors = validateURL(text, toCase);
+
+    if (errors.length) {
+      res.statusCode = 400;
```
33Orange, line 18: `If something is not correct, you should respond with 400 status, Bad request statusText`

Check the task requirements

33Orange (changes requested):
Almost done

PR #333, opened 2023-07-19
Solution

amashutikov:


src/handleErrors.js
```diff
@@ -0,0 +1,33 @@
...
+      { message: errorMessages.missedText },
+      { message: errorMessages.notListedCase },
+    );
+  } else if (!textToConvert) {
+    errors.push({ message: errorMessages.missedText });
+  } else if (!caseName) {
+    errors.push({ message: errorMessages.missedCase });
+  } else if (!caseNames.includes(caseName)) {
+    errors.push({ message: errorMessages.notListedCase });
+  }
```
33Orange, line 28: You have a lot of unnecessary checks.
You need only 3: 
`!textToConvert`, `!caseName`, `caseName && !caseNames.includes(caseName)`

And avoid using `else` and `else if`. Create all cases in a single `if` construction

src/createServer.js
```diff
@@ -1,3 +1,34 @@
...
+      res.end(JSON.stringify({ errors }));
+    } else {
+      const conversionResult = convertToCase(textToConvert, caseName);
+
+      conversionResult.targetCase = caseName;
+      conversionResult.originalText = textToConvert;
+
+      res.statusCode = 200;
+      res.end(JSON.stringify(conversionResult));
+    }
```
33Orange, line 28: Is `else` necessary here?
You have a case when you have the errors and default case.

 ```suggestion
    }
    const conversionResult = convertToCase(textToConvert, caseName);

    conversionResult.targetCase = caseName;
    conversionResult.originalText = textToConvert;

    res.statusCode = 200;
    res.end(JSON.stringify(conversionResult));
```

File: src/createServer.js [hunk 3]
```diff
@@ -1,3 +1,34 @@
...
+      res.statusMessage = 'Bad request';
+
+      res.end(JSON.stringify({ errors }));
+    } else {
+      const conversionResult = convertToCase(textToConvert, caseName);
+
+      conversionResult.targetCase = caseName;
+      conversionResult.originalText = textToConvert;
+
+      res.statusCode = 200;
```
33Orange, line 26: `You should respond with status 200 and OK status text.`

33Orange (changes requested):
Almost done

A few moments to fix

src/createServer.js
```diff
@@ -1,3 +1,43 @@
...
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const { originalCase, convertedText } = convertToCase(
+      textToConvert, caseName,
+    );
+
+    const result = {
```
silchyn, line 28: That's redundant

src/handleErrors.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable max-len */
+const caseNames = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const errorMessages = {
+  missedText: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  missedCase: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  notListedCase: `This case is not supported. Available cases: ${caseNames.join(', ')}.`,
+};
```
silchyn, line 7: Could be in a separate file like `constants.js`

src/handleErrors.js
```diff
@@ -0,0 +1,27 @@
...
+
+function handleErrors(textToConvert, caseName) {
```
silchyn, line 9: Do you actually handle errors here?

silchyn (approved):

PR #334, opened 2023-07-20
added solution

niaYaro:


src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+      errors.push({
+        message: '"toCase" query param is required. '
+          + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!acceptableCases.includes(targetCase)) {
+      errors.push({
+        message: 'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
```
33Orange, line 32: Create a separate helper which will return the array of errors.

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+          + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    if (!targetCase) {
+      errors.push({
+        message: '"toCase" query param is required. '
+          + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!acceptableCases.includes(targetCase)) {
```
33Orange, line 27: Avoid using `else`, `else if`

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+const { convertToCase } = require('./convertToCase');
+const acceptableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const createServer = () => (
+  http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const originalText = req.url.split('?')[0].slice(1);
+    const params = new URLSearchParams(req.url.split('?')[1]);
+    const targetCase = params.get('toCase');
```
33Orange, line 11: You can use `new URL` and then parse everything that you need from that URL.

```suggestion
    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
   
    const originalText = normalizedURL.pathname.slice(1);
    const targetCase = normalizedURL.searchParams.get('toCase');
```

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+          + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!acceptableCases.includes(targetCase)) {
+      errors.push({
+        message: 'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    if (errors.length > 0) {
```
33Orange, line 34:
```suggestion
    if (errors.length) {
```

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const convertedData = convertToCase(originalText, targetCase);
+    const { originalCase, convertedText } = convertedData;
```
33Orange, line 43:
```suggestion
    const { originalCase, convertedText } = convertToCase(originalText, targetCase);
```

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+    const result = {
+      originalCase,
+      targetCase,
+      originalText,
+      convertedText,
+    };
+
+    res.statusCode = 200;
+    res.end(JSON.stringify(result));
```
33Orange, line 52: `You should respond with status 200 and OK status text.`

Check the task requirements, please.

33Orange (changes requested):
Almost done

A few moments to fix

src/helper.js
```diff
@@ -0,0 +1,29 @@
+const getErrors = (givenText, givenCase) => {
+  const acceptableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
33Orange, line 2: Create this variable outside the function to avoid re-creating it every time when the function is executed

src/helper.js
```diff
@@ -0,0 +1,29 @@
...
+  const errors = [];
+
+  if (!givenText) {
+    errors.push({
+      message: 'Text to convert is required. '
+        + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
33Orange, line 8: You can create an object with all your default error messages and use them as reference

33Orange (changes requested):
Almost done

Just a few moments with constants

33Orange (approved):
Well done

PR #337, opened 2023-07-20
add solution

ivankhrsh:


src/createServer.js
```diff
@@ -1,3 +1,125 @@
...
+      res.write(JSON.stringify({
+        errors: [
+          { message: errorMessage.caseIsNotSupported },
+        ],
+      }));
+
+      res.end();
+
+      return;
+    }
```
33Orange, line 101: Create a separate function that will create an array of errors, push this array in some cases and return this array
After you will check if you have an error in the array and send all these errors like a response with code 400.

And you forget about the status text

```
const errors = validateQueryParams(text, targetCase);

if (errors.length) {
  res.statusCode = 400;
 ...
}
```

src/createServer.js
```diff
@@ -1,3 +1,125 @@
...
+
+    res.statusCode = 200;
```
33Orange, line 103: You should respond with status 200 and OK status text.

src/createServer.js
```diff
@@ -1,3 +1,125 @@
...
+
+    const result = {
+      originalCase,
+      targetCase,
+      originalText,
+      convertedText,
+    };
+
+    res.write(JSON.stringify(result));
+    res.end();
```
33Orange, line 117: You can combine this logic and provide JSON to the `end()` function

```suggestion
    res.end(JSON.stringify(result));
```

src/createServer.js
```diff
@@ -1,3 +1,125 @@
...
+    if (!supportedCases.includes(targetCase)) {
+      res.statusCode = 400;
+
+      res.write(JSON.stringify({
+        errors: [
+          { message: errorMessage.caseIsNotSupported },
+        ],
+      }));
+
+      res.end();
```
33Orange, line 98: Same

33Orange (changes requested):
Almost done

src/convertToCase/validateQueryParams.js
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  if (!text && !targetCase) {
+    errors.push(
+      { message: errorMessage.textIsRequired },
+      { message: errorMessage.caseIsIsRequired },
+    );
+
+    return errors;
+  }
```
33Orange, line 24: 2 same cases
```suggestion
  if (!text && !targetCase) {
    errors.push(
      { message: errorMessage.textIsRequired },
      { message: errorMessage.caseIsIsRequired },
    );

    return errors;
  }
```

src/convertToCase/validateQueryParams.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (!supportedCases.includes(targetCase)) {
+    errors.push(
+      { message: errorMessage.caseIsNotSupported },
+    );
+
+    return errors;
+  }
```
33Orange, line 57: 3 cases with same `!supportedCases.includes(targetCase)` logic ?

You can use only 3 if statements in this function.

```
if (!text) {
}

if(!targetCase) {
}

if(targetCase && !supportedCases.includes(targetCase)) {
}
```

33Orange (changes requested):
Almost done

PR #338, opened 2023-07-20
Develop

RoksolanaTylipska:


src/validation.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable max-len */
+
+const validCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+function validation(originalText, targetCase) {
+  const errors = [];
+
+  if (!originalText) {
+    errors.push({ message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
```
33Orange, line 9: You can create an object with all your default error messages and use them as reference

33Orange (approved):
Good job

PR #339, opened 2023-07-20

panch1811:


src/createServer.js
```diff
@@ -1,3 +1,115 @@
...
+      res.write(JSON.stringify({
+        errors: [
+          { message: errorMessage.unavailableCase },
+        ],
+      }));
+
+      res.end();
+
+      return;
+    }
```
33Orange, line 91: Too much-repeated logic. You send error responses in every case separately.

Let's create a separate function that will create an array of errors and check taken arguments for errors and return the array.

After that, you can use this array to send the response.

```
if (errors.length) {
  res.statusCode = 400;

  res.write(JSON.stringify(errors));
}
```

And you forgot about the status text.

src/createServer.js
```diff
@@ -1,3 +1,115 @@
...
+
+    const result = {
+      originalCase,
+      targetCase,
+      originalText,
+      convertedText,
+    };
+
+    res.write(JSON.stringify(result));
+    res.end();
```
33Orange, line 107: You can provide a response JSON to the `end()` argument

```suggestion
    res.end(JSON.stringify(result));
```

src/createServer.js
```diff
@@ -1,3 +1,115 @@
...
+    if (!availableCases.includes(targetCase)) {
+      res.statusCode = 400;
+
+      res.write(JSON.stringify({
+        errors: [
+          { message: errorMessage.unavailableCase },
+        ],
+      }));
+
+      res.end();
```
33Orange, line 88: Same here

src/createServer.js
```diff
@@ -1,3 +1,115 @@
...
+const { convertToCase } = require('./convertToCase/convertToCase');
+const http = require('http');
+
+const availableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const errorMessage = {
+  missingText: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  missingCase: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  unavailableCase: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+};
```
33Orange, line 11: Move this variables to a separate file

33Orange (changes requested):
Almost done

src/createServer.js
```diff
@@ -1,3 +1,55 @@
...
+
+    if (!originalText) {
+      errors.push({ message: errorMessage.missingText });
+    }
+
+    if (!targetCase) {
+      errors.push({ message: errorMessage.missingCase });
+    } else if (!availableCases.includes(targetCase)) {
+      errors.push({ message: errorMessage.unavailableCase });
+    }
```
33Orange, line 25: Rewrite this logic without `else` of `else if`

src/utils.js
```diff
@@ -0,0 +1,21 @@
...
+  missingCase: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  unavailableCase: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+};
+
+function createErrorResponse(res, statusCode, errors) {
+  res.statusCode = statusCode;
+  res.statusMessage = 'Bad Request';
+  res.setHeader('Content-Type', 'application/json');
+  res.end(JSON.stringify({ errors }));
+}
```
33Orange, line 15: You can use this function in the successful cases too

```suggestion
function sendResponse(res, statusCode, statusText, payload) {
  res.statusCode = statusCode;
  res.statusMessage = statusText;
  res.setHeader('Content-Type', 'application/json');
  res.end(JSON.stringify(payload));
}
```

src/createServer.js
33Orange, line 25: Let's move this logic to a separate function

33Orange (changes requested):

src/createServer.js
```diff
@@ -1,3 +1,63 @@
...
+  if (!targetCase) {
+    errors.push({ message: errorMessage.missingCase });
+  }
+
+  if (targetCase && !availableCases.includes(targetCase)) {
+    errors.push({ message: errorMessage.unavailableCase });
+  }
+
+  return errors;
+};
```
33Orange, line 25: Move it to a separate file

33Orange (approved):
Gj

PR #341, opened 2023-07-20
solution

thienlaoo:


src/errorhandler.js
```diff
@@ -0,0 +1,18 @@
...
+const isSearchParamsValid = (textInput, caseInput) => {
+  let error = '';
+
+  if (!textInput) {
+    error = 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+  } else if (!caseInput) {
+    error = '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+  } else if (!cases.includes(caseInput)) {
+    error = `This case is not supported. Available cases: ${cases}`;
+  }
```
33Orange, line 13: Let's remove an `else` and an `else if` constructions, and use only `if()`.
You need to create an array of errors and push this array in some cases. 
In this situation, we will have only 1 error, but we may not have both values


```
if (condition) {
    
}

if (condition) {
  
}
```

src/errorhandler.js
```diff
@@ -0,0 +1,18 @@
...
+const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const isSearchParamsValid = (textInput, caseInput) => {
+  let error = '';
+
+  if (!textInput) {
+    error = 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+  } else if (!caseInput) {
+    error = '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+  } else if (!cases.includes(caseInput)) {
```
33Orange, line 11:
```suggestion
  } else if (caseInput && !cases.includes(caseInput)) {
```

33Orange (changes requested):
Almost done

src/createServer.js
```diff
@@ -1,3 +1,38 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { isSearchParamsValid } = require('./errorhandler');
+
+const createServer = http.createServer((req, res) => {
```
33Orange, line 5: createServer should be a function

```suggestion
const createServer = () => http.createServer((req, res) => {
```

33Orange (changes requested):

PR #343, opened 2023-07-21
Solved

olehmahera:


src/createServer.js
```diff
@@ -1,3 +1,43 @@
 // Write code here
 // Also, you can create additional files in the src folder
 // and import (require) them here
```
33Orange, line 3: Remove redundant comments

33Orange (approved):
Good job

PR #347, opened 2023-07-22
solution

ghostargg:


src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+    if (errors.length) {
+      res.statusCode = 400;
+
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const { convertedText, originalCase } = convertToCase(words, newCase);
+
```
d-wasilewski, line 43: You are not handling the case if the 'convertToCase' function fails (throws an error). You can wrap this call in a try/catch block

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+    }
+
+    if (!newCase) {
+      errors.push({ message: '"toCase" query param is required. Correct '
+        + 'request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
+    }
+
+    if (!(caseNames.includes(newCase)) && newCase?.length > 1) {
+      errors.push({ message: 'This case is not supported. Available cases: '
+          + 'SNAKE, KEBAB, CAMEL, PASCAL, UPPER.' });
```
d-wasilewski, line 31: It will be easier to handle if you add/remove one
```suggestion
          + caseNames.join(', ')});
```

d-wasilewski (approved):

PR #348, opened 2023-07-23
add solution

PanyaPrimal:


src/createServer.js
```diff
@@ -1,3 +1,47 @@
...
+
+  return response.end(JSON.stringify(data));
+};
+
+const createServer = () => {
+  const server = http.createServer((request, response) => {
+    const [text, toCase] = getParams(request);
+    const errors = validate(text, toCase);
+
+    if (errors.length > 0) {
```
33Orange, line 26:
```suggestion
    if (errors.length) {
```

src/createServer.js
```diff
@@ -1,3 +1,47 @@
...
+  const text = pathname.slice(1);
+  const toCase = searchParams.get('toCase');
+
+  return [text, toCase];
+};
+
+const sendResponse = (response, statusCode, data) => {
+  response.setHeader('Content-Type', 'application/json');
+  response.statusCode = statusCode;
+
```
33Orange, line 17: You forgot about the status text in the responses

`You should respond with status 200 and OK status text.`

33Orange (changes requested):
Almost done

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+    case 400:
+      response.statusMessage = 'Bad Request';
+      break;
+    default:
+      response.statusMessage = '';
+      break;
+  }
+
+  return response.end(JSON.stringify(data));
+};
```
33Orange, line 31: Move this functions to a separate file

33Orange (approved):
Well done

PR #350, opened 2023-07-23
add solution

ulositska:


src/checkUrlParams.js
```diff
@@ -0,0 +1,24 @@
+/* eslint-disable max-len */
+const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+function checkUrlParams(textToConvert, toCase) {
+  const errors = [];
+
+  if (!textToConvert) {
+    errors.push({ message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
```
33Orange, line 8: You can create an object with all your default error messages and use them as reference

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+    const toCase = urlParams.searchParams.get('toCase');
+
+    const errors = checkUrlParams(textToConvert, toCase);
+
+    if (errors.length > 0) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+      res.end(JSON.stringify({ errors }));
+    } else {
+      const result = convertToCase(textToConvert, toCase);
```
33Orange, line 20: Is it necessary to use `else` here? 
You can add the return to the if statement and use code under as a default

```suggestion
    if (errors.length > 0) {
      res.statusCode = 400;
      res.statusMessage = 'Bad request';
      res.end(JSON.stringify({ errors }));
      return;
    } 

    const result = convertToCase(textToConvert, toCase);
```

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const urlParams = new URL(req.url, `http://${req.headers.host}`);
+    const textToConvert = urlParams.pathname.slice(1);
+    const toCase = urlParams.searchParams.get('toCase');
+
+    const errors = checkUrlParams(textToConvert, toCase);
+
+    if (errors.length > 0) {
```
33Orange, line 15:
```suggestion
    if (errors.length) {
```

src/createServer.js
33Orange, line 20:
```suggestion
      const { originalCase, convertedText } = convertToCase(textToConvert, toCase);
```

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+      res.statusMessage = 'Bad request';
+      res.end(JSON.stringify({ errors }));
+    } else {
+      const result = convertToCase(textToConvert, toCase);
+      const responsePayload = {
+        originalCase: result.originalCase,
+        targetCase: toCase,
+        originalText: textToConvert,
+        convertedText: result.convertedText,
+      };
```
33Orange, line 26:
```suggestion
      const responsePayload = {
        originalCase,
        targetCase: toCase,
        originalText: textToConvert,
        convertedText,
      };
```

33Orange (changes requested):
Almost done

PR #353, opened 2023-07-26
Server created

Yaroslav-Kononenko:


src/isError.js
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * @param {CaseName | undefined} param
+ * @param {string} pathname
+ *
+ * @returns {Array<{ message: string }> | undefined}
+ */
+
+function isError(param, pathname) {
+  const errors = [];
+  const isAvailable = ['SNAKE' , 'KEBAB' , 'CAMEL' , 'PASCAL' , 'UPPER'].some((toCase) => toCase === param);
```
Anna-Tkachenko, line 14: Let's create separate constant with available cases values

src/isError.js
```diff
@@ -0,0 +1,37 @@
...
+
+  if(!pathname) {
+    errors.push({
+      message: `Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".`
```
Anna-Tkachenko, line 18: Will be great to move the messages for the different errors to the constants too

Anna-Tkachenko (approved):

PR #355, opened 2023-08-01
add task solution

ollavka:


src/createServer.js
```diff
@@ -1,3 +1,73 @@
...
+      || caseNameIsNotSupported;
+
+    if (hasError) {
+      const errorMessages = {
+        errors: [],
+      };
+
+      if (textIsMissing) {
+        errorMessages.errors.push({
+          message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
And678, line 35: Move those messages outside of the main code, to constants.js file.

src/createServer.js
```diff
@@ -1,3 +1,73 @@
...
+    const data = {
+      originalCase,
+      convertedText,
+      targetCase,
+      originalText,
+    };
+
+    res.statusCode = 200;
+    res.statusMessage = 'OK';
+    res.end(JSON.stringify(data));
```
And678, line 67: From [Express docs](http://expressjs.com/en/4x/api.html#res.end) about .end: `Use to quickly end the response without any data. If you need to respond with data, instead use methods such as res.send() and res.json().` 
```suggestion
    res
      .status(200)
      .send(JSON.stringify(data));
```

And678 (approved):
Good work, but check my comments

src/createServer.js
And678: Oops, wrong homework :smiley:

PR #356, opened 2023-08-02
Solution

Zibi95:


src/createServer.js
```diff
@@ -1,3 +1,83 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase.js');
+
+const validateInput = (originalText, targetCase) => {
```
Radoslaw-Czerniawski, line 5: Generally when working with servers it's very important to split most of the stuff into separate modules / files. It decouples your applications, enables easier modifications and separation of the business logic.

Ideally you would create a server directory exporting helpers / modules.
image
image

Radoslaw-Czerniawski: image

Radoslaw-Czerniawski: image

Radoslaw-Czerniawski: image

Radoslaw-Czerniawski: image

Radoslaw-Czerniawski: And finally the implementation of the createServer
image

Radoslaw-Czerniawski (changes requested):
I did some hardcore module splitting, no need to implement exactly same solution (however it definitely would work :D) try to do some refactoring.

src/createServer.js
Zibi95: I'm wondering, why should this be so explicit?
image
If I just send statusCode, the default message would be the same.

darokrk (approved):
Good job 👏🏽 🏅

PR #357, opened 2023-08-06
Solution

MihailKostov:


src/isError.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable max-len */
+const availableCases = ['SNAKE' | 'KEBAB' | 'CAMEL' | 'PASCAL' | 'UPPER'];
```
VitaliyBodnar, line 2: The elements of the 'availableCases' array need to be comma-separated strings. In the current format with the '|' operator, it's not creating an array of strings but rather a bitwise OR operation.

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+/* eslint-disable max-len */
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+const { isError } = require('./isError');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const normalizedURL = new URL(req.url, `http://localhost${req.headers.host}`);
```
VitaliyBodnar, line 10: Possible error in concatenating request headers host with 'http://localhost'. The host should be used as is without adding 'http://localhost'.

src/createServer.js
```diff
@@ -1,3 +1,40 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
```
VitaliyBodnar, line 1: Avoid disabling eslint rules without a valid reason. It is better to adjust the code to meet the eslint rules.

tests/createServer.test.js
```diff
@@ -1,206 +1,73 @@
...
-        resolve({ res, body });
-      });
-    })
-      .on('error', (err) => {
-        reject(err);
-      });
-  });
-}
+    const normalizedURL = new URL(
+      req.url, `http://localhost${req.headers.host}`,
```
VitaliyBodnar, line 14: The URL constructor is not properly initiated. The base URL should be `http://${req.headers.host}`

tests/createServer.test.js
```diff
@@ -1,206 +1,73 @@
...
-const { Server, get } = require('http');
+const http = require('http');
+const { convertToCase } = require('../src/convertToCase');
 
-function request(url = '/') {
-  return new Promise((resolve, reject) => {
-    get('http://localhost:5701' + url, (res) => {
-      let body = '';
+const createServer = () => {
+  const availableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
VitaliyBodnar, line 8: The variable `availableCases` should be declared as a constant outside the `createServer` function, as it doesn't change throughout the function execution.

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+    const data = {
+      originalCase,
+      convertedText,
+      targetCase,
+      originalText,
+    };
+
+    res.statusCode = 200;
+    res.statusMessage = 'OK';
+    res.end(JSON.stringify(data));
```
VitaliyBodnar, line 34: can be less
please check other places
```suggestion
    res.status(200).send(data);
```

VitaliyBodnar (changes requested):
Almost done!

src/createServer.js
MihailKostov: there are an error: res.status is not a function

PR #364, opened 2023-08-14
solution

cebanoleksandr:


src/createServer.js
```diff
@@ -1,3 +1,57 @@
...
+
+function createServer() {
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+  const server = http.createServer((req, res) => {
+    const [text, queryString] = req.url.slice(1).split('?');
+    const queryParams = new URLSearchParams(queryString);
+    const toCase = queryParams.get('toCase');
+
+    if (!text || !toCase || !cases.includes(toCase)) {
```
And678, line 13: Instead of duplicating check like `!cases.includes(toCase)`, just check for `errors.length` afterwards
```suggestion
```

src/createServer.js
```diff
@@ -1,3 +1,57 @@
...
+
+      if (!text) {
+        errors.push({ message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
+      }
+
+      if (!toCase) {
+        errors.push({ message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' });
+      } else if (!cases.includes(toCase)) {
+        errors.push({ message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.' });
+      }
```
And678, line 24: This would be better in a separate validation file

And678 (changes requested):

anstsot (approved):

PR #371, opened 2023-08-15
solution

AnatoliiLykhohodenko:


src/createServer.js
```diff
@@ -1,3 +1,68 @@
...
+function isValidCase(someCase) {
+  return cases.includes(someCase);
+}
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-type', 'application/json');
+
+    const url = new URL(req.url, `http://${req.headers.host}`);
+    const Case = url.searchParams.get('toCase');
```
And678, line 18: Don't start variable names with capital letters, they are reserved for class names, etc.
```suggestion
    const сase = url.searchParams.get('toCase');
```

src/createServer.js
```diff
@@ -1,3 +1,68 @@
...
+    const recievedText = url.pathname.slice(1);
+
+    const isError = !Case || !recievedText || !isValidCase(Case);
+
+    if (isError) {
```
And678, line 23: Instead of doing this, just check for `errorResponse.errors.length !== 0` to check if we have any errors (after your `if`s)
```suggestion
```

And678 (changes requested):
Good job, but can be even better :)

YegorKorenkov (approved):
Good job!

PR #372, opened 2023-08-15
Solution

perfectAcedia:


src/createServer.js
```diff
@@ -1,3 +1,73 @@
...
+
+    const res = {
+      originalCase: convertedInput.originalCase,
+      targetCase: toCase,
+      originalText: textToConvert,
+      convertedText: convertedInput.convertedText,
+    };
+
+    response.writeHead(200, { 'Content-Type': 'application/json' });
+    response.end(JSON.stringify(res));
```
And678, line 67: Can be chained here
```suggestion
    response.writeHead(200, { 'Content-Type': 'application/json' })
      .end(JSON.stringify(res));
```

src/createServer.js
```diff
@@ -1,3 +1,73 @@
...
+      });
+    // eslint-disable-next-line no-prototype-builtins
+    } else if (!caseOptions.hasOwnProperty(toCase)) {
+      errors.push({
+        message:
+          'This case is not supported. '
+          + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
```
And678, line 46: Better have this logic in separate file

And678 (approved):
Good, but check my comment :heavy_check_mark:

PR #375, opened 2023-08-15
Add task solution

vladmakohon:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+        convertedText,
+      } = convertToCase(originalText, targetCase);
+
+      responseData = {
+        originalCase,
+        targetCase,
+        originalText,
+        convertedText,
+      };
+    }
```
And678, line 55: Use of `let` is avoidable here
```suggestion

    response.setHeader('content-type', 'application/json');

    const errors = [];

    if (originalText.length === 0) {
      errors.push({
        message: ERROR_MESSAGES.noTextToConvert,
      });
    }

    if (!targetCase) {
      errors.push({
        message: ERROR_MESSAGES.noCaseParam,
      });
    } else if (!SUPPORTED_CASES.includes(targetCase)) {
      errors.push({
        message: ERROR_MESSAGES.notSupportedCase,
      });
    }

    if (errors.length) {
      const responseData = {
        errors,
      };

      response.statusCode = 400;
      response.statusMessage = 'Bad request';
    } else {
      const {
        originalCase,
        convertedText,
      } = convertToCase(originalText, targetCase);

      const responseData = {
        originalCase,
        targetCase,
        originalText,
        convertedText,
      };
    }
```

src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+
+    if (!targetCase) {
+      errors.push({
+        message: ERROR_MESSAGES.noCaseParam,
+      });
+    } else if (!SUPPORTED_CASES.includes(targetCase)) {
+      errors.push({
+        message: ERROR_MESSAGES.notSupportedCase,
+      });
+    }
```
And678, line 34: This logic would look better in separate file for validation

And678 (approved):
Good, but check my comment below

src/createServer.js
vladmakohon: Thanks, but I'm not sure if this will work. If we declare `responseData` only inside `if...else` statement, then how are we going to pass this object to `response.end()` method, which is called outside?

PR #380, opened 2023-08-16
add task solution

irynabolbat:


src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+
+      if (!getCase) {
+        errors.push({
+          message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+        });
+      } else if (!supportedCases.includes(getCase)) {
+        errors.push({
+          message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+        });
+      }
```
And678, line 32: Better to move validation to another file.

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+  const server = http.createServer((req, res) => {
+    const normalizeUrl = new URL(req.url, `http://${req.headers.host}`);
+
+    const text = normalizeUrl.pathname.slice(1);
+    const getCase = normalizeUrl.searchParams.get('toCase');
+
+    if (!text || !getCase || !supportedCases.includes(getCase)) {
+      const errors = [];
+
+      if (!text) {
```
And678, line 18: Instead of checking for `!text` (and others) twice, just check for `errors.length !== 0` afterwards
```suggestion
      const errors = [];

      if (!text) {
```

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+
+    const response = {
+      originalCase: convertedText.originalCase,
+      targetCase: getCase,
+      originalText: text,
+      convertedText: convertedText.convertedText,
+    };
+
+    res.writeHead(200, { 'Content-Type': 'application/json' });
+    res.end(JSON.stringify(response));
```
And678, line 50: This can be chained
```suggestion
    res.writeHead(200, { 'Content-Type': 'application/json' })
      .end(JSON.stringify(response));
```

And678 (changes requested):
Good job, but can be better :)

anstsot (approved):

PR #382, opened 2023-08-16
add task solution

AlyonaSarapina:


src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/index');
+
+const caseOptions = ['SNAKE', 'KEBAB', 'UPPER', 'CAMEL', 'PASCAL'];
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const [text, toCase] = req.url.substring(1).split('?toCase=');
```
And678, line 8: Cool idea to split by `?toCase=` :fire: 
```suggestion
    const [text, toCase] = req.url.substring(1).split('?toCase=');
```

src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+
+    const response = {
+      originalCase: result.originalCase,
+      targetCase: toCase,
+      originalText: text,
+      convertedText: result.convertedText,
+    }
+
+    res.writeHead(200, { 'Content-Type': 'application/json' });
+    res.end(JSON.stringify(response));
```
And678, line 60: `writeHead` returns `res` back so this can be chained
```suggestion
    res.writeHead(200, { 'Content-Type': 'application/json' })
      .end(JSON.stringify(response));
```

And678 (approved):
:+1:

PR #388, opened 2023-08-16
Dev

OKhalanskyi:


.idea/jsLibraryMappings.xml
And678, line 1: Don't push `.idea` to solution, this is your personal IDE's project config

And678 (approved):
:white_check_mark:

PR #390, opened 2023-08-16
task solution

AnastasiiaKhemych:


.idea/inspectionProfiles/Project_Default.xml
And678, line 1: Don't do `git add .` or `git add -A`, as you pushed a lot of your personal file in `.idea` folder. Just push only the files that are needed

And678 (approved):
Approve, but check my comment

PR #392, opened 2023-08-16
added task solution

tetianakostiuk1:


src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+
+function createServer() {
+  return http.createServer((req, res) => {
+    const url = new URL(`http://${req.headers.host}${req.url}`);
+    const targetCase = url.searchParams.get('toCase');
+    const originalText = url.pathname.slice(1);
+
+    const isError = !originalText
+      || !targetCase
+      || !availableCases.includes(targetCase);
```
And678, line 14: You already have these checks below, don't duplicate them, just check for errors.length to send an error
```suggestion
```

And678 (changes requested):
Nice work, only one case of duplication needs to be removed

anstsot (approved):

PR #393, opened 2023-08-16
Solution

melnyksergey7:


src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+    originalText,
+    convertedText,
+  ) {
+    return {
+      originalCase,
+      targetCase,
+      originalText,
+      convertedText,
+    };
+  };
```
And678, line 31: As i think, this helps to ensure that the response will have valid structure, but it's too redundant, you can just create this object in-place.
```suggestion
```

src/validationCase.js
And678, line 1: Nice idea having this in a separate file :fire:

And678 (approved):
Very good, but check my comment

PR #399, opened 2023-08-17
add task solution

EvgeniyChumak:


src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+      response.body = {
+        originalText: convertedText,
+        targetCase: targetCase.toUpperCase(),
+        convertedText: result.convertedText,
+        originalCase: result.originalCase,
+      };
+    }
+
+    res.writeHead(response.status, response.headers);
+    res.end(JSON.stringify(response.body));
```
anstsot, line 53: can be chained
```suggestion
    res.writeHead(response.status, response.headers)
      .end(JSON.stringify(response.body));
```

anstsot (approved):
Good work, approving✨

PR #403, opened 2023-08-17
add solution

pavelkuraksin:


src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+      if (!CASE_PARAMS.includes(toCase) && toCase) {
+        throwErrorMessage.errors.push({
+          message: errorMessages.caseNotFined,
+        });
+      }
+
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+      res.end(JSON.stringify(throwErrorMessage));
+    } else {
```
anstsot, line 48: you can check the `throwErrorMessage.error.length` instead of `if` on line 26

anstsot (approved):

PR #405, opened 2023-08-20

NataliGru:


src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+        errorMessages.push({
+          message: `This case is not supported. Available cases: ${SUPPORTED_CASES.join(', ')}.`,
+        });
+      }
+
+      res.writeHead(400, 'Bad request');
+
+      res.end(JSON.stringify({
+        errors: errorMessages,
+      }));
```
anstsot, line 46: just check `errorMessages.length`
```suggestion

    if (errorMessages.length > 0) {
      res.writeHead(400, 'Bad request')
        .end(JSON.stringify({ errors: errorMessages }));
      
      return;
    }
```

anstsot (approved):
Good work, approving✨

PR #406, opened 2023-08-20
The task has been completed

bojkovladislav:


src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+      }
+
+      if (availableCases.every(availableCase => availableCase !== targetCase && targetCase)) {
+        errors.push({
+          "message": `This case is not supported. Available cases: ${availableCases.join(', ')}.`
+        })
+      }
+
+        res.writeHead(400, 'Bad request');
+        res.statusCode = 400;
```
loralevitska, line 42: There's no need to set the statusCode after calling res.writeHead. The statusCode is already set by res.writeHead.

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+function createServer() {
+  return http.createServer((req, res) => {
+    res.setHeader('Content-type', 'application/json');
+
+    const newUrl = new URL(req.url, `http://${req.headers.host}`);
+
+    const targetCase = newUrl.searchParams.get('toCase');
+    const textToConvert = newUrl.pathname.slice(1);
+
+    const isError = !textToConvert
```
loralevitska, line 16: The condition for checking isError could be simplified. Instead of using every and !==, you can use includes to check if targetCase is in the availableCases array.

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+        })
+      }
+
+      if (!targetCase) {
+        errors.push({
+          "message": '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+        })
+      }
+
+      if (availableCases.every(availableCase => availableCase !== targetCase && targetCase)) {
```
loralevitska, line 35: The condition for checking if the targetCase is in availableCases is unnecessary. Just check if the targetCase is not in availableCases.

src/createServer.js
```diff
@@ -1,3 +1,60 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { URL } = require('url')
+const {convertToCase } = require('./convertToCase/convertToCase');
```
loralevitska, line 3:
```suggestion
const { convertToCase } = require('./convertToCase/convertToCase');
```

loralevitska (approved):
Well done)

PR #408, opened 2023-08-21
add task solution

parmenid89:


src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+
+      res.statusCode = 404;
+      res.statusMessage = 'Bad request';
+
+      res.end(
+        JSON.stringify({ errors }),
+      );
+
+      return;
+    }
```
anstsot, line 45:
```suggestion
    if (errors.length > 0) {
      res.statusCode = 404;
      res.statusMessage = 'Bad request';

      res.end(JSON.stringify({ errors }));

      return;
    }
```

src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const originalText = normalizedURL.pathname.slice(1);
+    const targetCase = normalizedURL.searchParams.get('toCase');
+
+    const textIsMissing = !originalText;
+    const toCaseIsMissing = !targetCase;
+    const caseNameIsMissing = !(caseName.includes(targetCase));
+
+    if (textIsMissing || toCaseIsMissing || caseNameIsMissing) {
+      const { errorMessages } = require('./errorMessages');
```
anstsot, line 22: better to require on the top of file

anstsot (approved):
Good work, approving✨

PR #411, opened 2023-08-22
Solution

Ilia991:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
 // Write code here
 // Also, you can create additional files in the src folder
 // and import (require) them here
+const http = require('http');
+const fs = require('fs');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  return http.createServer((req, res) => {
+    const fileName = req.url.slice(1);
```
lerastarynets, line 10: This line slices the URL at the first character which is not necessary. You can directly get the pathname from urlParams which you created on the next line.

src/createServer.js
```diff
@@ -1,3 +1,42 @@
 // Write code here
 // Also, you can create additional files in the src folder
 // and import (require) them here
+const http = require('http');
+const fs = require('fs');
```
lerastarynets, line 5: The 'fs' module is required but never used. It's best practice to remove unused imports to keep the code clean.

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  return http.createServer((req, res) => {
+    const fileName = req.url.slice(1);
+    const urlParams = new URL(req.url, `http://${req.headers.host}`);
+    const pathName = urlParams.pathname;
+
+    if (pathName === '/favicon.ico') {
```
lerastarynets, line 14: If the pathName is '/favicon.ico', the function returns without sending any response. It's better to send a 404 status code and a message indicating that the favicon is not found.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

maxim2310 (approved):
Well done!!!

PR #419, opened 2023-08-26
solution

Kinderwagen:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
-// and import (require) them here
+'use strict';
+
+const http = require('http');
+const { checkErrors } = require('./validation');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normalizedUrl = new URL(req.url, 'http://localhost:5700');
```
lerastarynets, line 9: Hardcoding a host name and port number can be limiting and might not work in all environments. 
```suggestion
    const normalizedUrl = new URL(req.url, `http:${req.headers.host}`);
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #423, opened 2023-08-31
finished task  node_case-transformer-server

Angelion541:


src/createServer.js
```diff
@@ -1,3 +1,75 @@
...
+
+      Object.assign(CONVERTATION_RESULT, CONVERTATION_ORIGINAL_VALUES);
+
+      Object.assign(RESULT_DATA, CONVERTATION_RESULT);
+      res.statusCode = 200;
+    } else {
+      res.statusCode = 400;
+    }
+
+    res.statusCode = RESULT_DATA.errors ? 400 : 200;
```
polosanya, line 69: What's the purpose of this line? You've already set the statusCode before

src/createServer.js
```diff
@@ -1,3 +1,75 @@
...
+    res.statusMessage = RESULT_DATA.errors ? 'Bad request' : 'OK';
```
polosanya, line 70: Same here. You can set it before

src/createServer.js
```diff
@@ -1,3 +1,75 @@
...
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable operator-linebreak */
+function createServer() {
+  const http = require('http');
+
+  return http.createServer((req, res) => {
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const RESULT_DATA = {};
+    const TEXT = normalizedURL.pathname.slice(1);
```
polosanya, line 8:
```suggestion
    const resultData = {};
    const text = normalizedURL.pathname.slice(1);
```
It's better to use camelCase for calculated variables

src/createServer.js
```diff
@@ -1,3 +1,75 @@
...
+      } else {
+        RESULT_DATA.errors = [NO_CASE_MESSAGE];
+      }
+    }
+
+    if (CASE_NAME && !CASE_VARIANTS.includes(CASE_NAME)) {
+      const INVALID_CASE_MESSAGE = {
+        message:
+          'This case is not supported. ' +
+          'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
polosanya, line 42: You can use CASE_VARIANTS instead of the hardcoded values

polosanya (changes requested):
Good job! Let's improve your code a bit)

polosanya (approved):
Good job!

PR #425, opened 2023-09-14
the task was done

pushkarskiyrodion:


src/validation.js
```diff
@@ -0,0 +1,30 @@
+const validation = (text, toCase) => {
```
mykhalenych, line 1: The function name 'validation' is not a verb or verb phrase. Consider renaming it to something that better describes its action like 'validateTextAndCase'.

src/request.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable no-console */
+const http = require('http');
+
+const BASE = 'http://localhost:5700';
```
mykhalenych, line 4: The BASE variable is hardcoded. It would be better to store this in a configuration file or environment variable for better flexibility.

src/request.js
```diff
@@ -0,0 +1,12 @@
...
+
+const href = BASE + '/hello-world?toCase=CAMEL';
+
+const request = http.request(href, (res) => {
```
mykhalenych, line 8: The response from the request is being logged to the console. This does not meet the requirement to respond with the converted text in JSON format. You should parse the response and return it in JSON format.

src/request.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable no-console */
+const http = require('http');
+
+const BASE = 'http://localhost:5700';
+
+const href = BASE + '/hello-world?toCase=CAMEL';
```
mykhalenych, line 6: The 'href' variable is hardcoded. This doesn't comply with the task requirement to accept requests in the format '/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>'. The parameters should not be hardcoded and should be designed to accept dynamic input.

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+    const normalizedURL = new URL(req.url, `http://${req.headers.host}`);
+    const text = normalizedURL.pathname.replace('/', '');
+    const caseName = normalizedURL.searchParams.get('toCase');
+
+    res.setHeader('Content-type', 'application/json');
+
+    const errors = validation(text, caseName);
+
+    if (errors.length) {
+      res.status = 400;
```
mykhalenych, line 16: The correct property to set the status code is `statusCode`, not `status`. Correct this to `res.statusCode = 400;`

src/createServer.js
```diff
@@ -1,3 +1,35 @@
...
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const converted = convertToCase(text, caseName.toUpperCase());
```
mykhalenych, line 22: The `convertToCase` function is expected to handle case conversion but here it is receiving an already uppercased `caseName`. This might not be correct depending on the implementation of `convertToCase`. Ensure that `convertToCase` expects an uppercased case name, or pass the original `caseName` to it.

mykhalenych (changes requested):

loralevitska (approved):
Well done)

PR #427, opened 2023-09-19
Solution

artyomwhite:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
-// and import (require) them here
+'use strict';
+
+const http = require('http');
+const { checkErrors } = require('./validation');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const normalizedUrl = new URL(req.url, 'http://localhost:5700');
```
lerastarynets, line 9: The hardcoded localhost URL in URL constructor might be a problem when deploying on a different server or port. It would be better to take hostname and port from the request headers or use environment variables.
```suggestion
    const normalizedUrl = new URL(req.url, `http:${req.headers.host}`);
```

lerastarynets (approved):
Great job overall! Left one tiny suggestion for the improvement

PR #430, opened 2023-10-06
Solution

Viktorianeimesh:


maxim2310 (changes requested):
Good work, but would be better to move all validation to separate func

nazarmatsevych (approved):
Great work!

PR #439, opened 2023-10-10
Develop

ivannaromanovych:


src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+      try {
+        convertToCase(originalText, targetCase);
+      } catch (err) {
+        erorr.errors.push({
+          message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.'
+        });
+      }
+    }
+
+    if (erorr.errors.length > 0) {
```
polinavafik, line 38:
```suggestion
    if (erorr.errors.length) {
```
it'll work the same

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+      res.end(JSON.stringify(erorr));
+    } else {
```
polinavafik, line 42: i think if you remove else it would still work the same (but then you need to add `return` to your if block, so everything else after if wont be executed)

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+      res.statusCode = 200;
+      res.statusMessage = 'OK';
+      const { originalCase, convertedText } = convertToCase(originalText, targetCase);
```
polinavafik, line 45: smart!

polinavafik (approved):
Amazing!! 🎀✨👍

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+const { convertToCase } = require('./convertToCase');
+const { getParams } = require('./getParams');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const { errors, targetCase, originalText } = getParams(req.url);
+
+    if (errors.length > 0) {
```
roman-mirzoian, line 11:
```suggestion
    if (errors.length) {
```

src/createServer.js
roman-mirzoian: Good point!

roman-mirzoian (approved):
Well done, but pay attention to the comments.

PR #440, opened 2023-10-10
create working http server

rostyslav48:


src/createServer.js
```diff
@@ -1,3 +1,50 @@
...
+      originalText: textToConvert,
+      targetCase: style,
+    };
+
+    res.end(JSON.stringify(preparedResponse));
+  });
+
+  // server.listen(PORT, () => {
+  //   // eslint-disable-next-line
+  //   console.log(`The server is running on ${serverLink}`);
```
cty2802, line 42: Don't forget to delete the comment.

cty2802 (approved):

src/createServer.js
rostyslav48: Oops👀

src/urlValidator.js
```diff
@@ -0,0 +1,37 @@
...
+    return [noStyle];
+  }
+
+  if (!textToConvert) {
+    return [noTextToConvert];
+  }
+
+  if (!AVALIABLE_STYLES.includes(style)) {
+    return [wrongStyle];
+  }
```
BudnikOleksii, line 32: You can create an errors array, push all errors that occur, and then return it from this function

src/urlValidator.js
```diff
@@ -0,0 +1,37 @@
...
+function urlValidator(textToConvert, style) {
+  const noStyle = {
+    message: `"toCase" query param is required. ${REQUEST_EXAMPLE}.`,
+  };
+  const noTextToConvert = {
+    message: `Text to convert is required. ${REQUEST_EXAMPLE}.`,
+  };
+  const wrongStyle = {
+    message: `This case is not supported. Available cases: ${AVALIABLE_STYLES.join(', ')}.`,
+  };
```
BudnikOleksii, line 12: You can create just one constant object with all errors messages and use it, also move it outside function

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

BudnikOleksii (approved):
🚀

PR #441, opened 2023-10-10
create server, implement error messages and case transform

serkrops:


src/constatnts.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable max-len */
+const STATIC_CASES = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
rostyslav48, line 2: Maybe it is better to rename this variable `POSSIBLE_STYLES` , now it is not clear which cases we are talking about

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+const {
+  NO_TEXT_ERROR_MESSAGE,
+  NO_CASE_ERROR_MESSAGE,
+  STATIC_CASES,
+  INVALID_CASE_ERROR_MESSAGE,
+} = require('./constatnts');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const url = new URL(req.url, 'http://localhost:5700');
```
rostyslav48, line 12: What would you do if the port is changed? It is a hardcoded value and needs to be fixed. You can get a port from res, here is an example    ` const serverLink = `http://${req.headers.host}`;`

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+    res.setHeader('Content-Type', 'application/json');
+
+    const textFromUrl = url.pathname.slice(1);
+    const caseFromUrl = url.searchParams.get('toCase');
+
+    const errors = {
+      errors: [],
+    };
+
+    if (textFromUrl === '') {
```
rostyslav48, line 23: It is better to move all these if statements to the separate component

src/createServer.js
```diff
@@ -1,3 +1,59 @@
...
+      const result = {
+        originalCase,
+        targetCase: caseFromUrl,
+        originalText: textFromUrl,
+        convertedText,
+      };
+
+      res.end(JSON.stringify(result));
+    } else {
+      res.end(JSON.stringify(errors));
```
rostyslav48, line 50: I cannot find where you add some code to your response headers.
You should respond with status 200 and `OK` status text.
If something is not correct, you should respond with 400 status, `Bad request` statusText and the next payload:
```
{
  "errors": [
    {
      "message": "<SPECIFIC MESSAGE TEXT HERE>"
    }
  ],
}
```

rostyslav48 (changes requested):
Well done👍
Although, there is one critical comment which needs to be fixed (about choosing the port). Other suggestions are on your consideration)

src/getErrors.js
```diff
@@ -0,0 +1,29 @@
...
+  NO_CASE_ERROR_MESSAGE,
+  INVALID_CASE_ERROR_MESSAGE,
+} = require('./constatnts');
+
+function getError(textFromUrl, caseFromUrl, ALLOWED_CASES) {
+  const errors = {
+    errors: [],
+  };
+
+  if (textFromUrl === '') {
```
roman-mirzoian, line 12:
```suggestion
  if (!textFromUrl) {
```

roman-mirzoian (approved):
Well done 👍

PR #445, opened 2023-10-10
add task solution

cty2802:


Wesses (approved):
good job!

src/createServer.js
```diff
@@ -1,3 +1,74 @@
...
+      'This case is not supported. '
+      + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+    );
+  }
+
+  return !errors.length
+    ? null
+    : {
+      errors: errors.map((errMessage) => ({ message: errMessage })),
+    };
```
roman-mirzoian, line 73:
```suggestion
  return errors.length ? { errors: errors.map(message => ({ message })) } : null;
```

roman-mirzoian (approved):
In small projects, it is better to keep all constants together.
It is also worth putting the error text in separate constants.
But overall, good job!

PR #446, opened 2023-10-10
add task solution

yar0man:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+    res.setHeader('Content-Type', 'application/json');
+
+    const originalText = myURL.pathname.slice(1);
+    const targetCase = myURL.searchParams.get('toCase');
+
+    const errors = handleErrors(originalText, targetCase);
+
+    if (errors.length) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
```
desevoker, line 18:
```suggestion
```
That's the default message for 400 status code, there is no need to set it explicitly.

src/createServer.js
```diff
@@ -1,3 +1,42 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { handleErrors } = require('./handleErrors');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const myURL = new URL(req.url, 'http://localhost:5700');
```
desevoker, line 7:
```suggestion
    const myURL = new URL(req.url, 'http://' + req.headers.host);
```

desevoker (approved):
Good job! 🚀

src/createServer.js
roman-mirzoian: Good point!

roman-mirzoian (changes requested):
Almost ready, pay attention to the comments of the previous review

Moroz-Dmytro (approved):

PR #465, opened 2023-10-11
created createServer function

edgergard:


dpidlutska (approved):
LGTM! 🚀

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+
+      return;
+    }
+
+    const {
+      originalCase,
+      convertedText
+    } = convertToCase(originalText, targetCase);
+
+    response.statusCode = 200;
```
BudnikOleksii, line 27: You forgot to set the statusText:
"You should respond with status 200 and OK status text."

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+    const requestURL = new URL(request.url, 'http://localhost:5700');
+    response.setHeader('Content-Type', 'application/json');
+
+    const originalText = requestURL.pathname.slice(1);
+    const targetCase = requestURL.searchParams.get('toCase');
+    const errors = getError(originalText, targetCase);
+
+    if (errors.errors.length) {
+      response.write(JSON.stringify(errors));
+      response.statusCode = 400;
```
BudnikOleksii, line 16: You forgot to set the statusText:
"If something is not correct, you should respond with 400 status, Bad request statusText and the next payload"

BudnikOleksii (changes requested):
Good job 👍 
Just fix one moment that I mentioned in the comments

Moroz-Dmytro (changes requested):

PR #466, opened 2023-10-11
Node case transformer server second commit

Sergey-Mironenko:


src/createServer.js
```diff
@@ -1,3 +1,70 @@
...
+// Write code here
+// Also, you can create additional files in the src folder
+// and import (require) them here
+
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+    const server = http.createServer((request, response) => {
+    const normalizedUrl = new URL(request.url, `http://${request.headers.url}`);
```
denys-danyliuk, line 10:
```suggestion
    const server = http.createServer((request, response) => {
      const normalizedUrl = new URL(request.url, `http://${request.headers.url}`);

```
nested code should not have same indentation

denys-danyliuk (approved):

PR #482, opened 2023-10-18
Solution

Nightycake:


Matylda-Lewandowska (approved):
Looks good to me! :)

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+  });
+
+  return server;
+};
+
+module.exports = {
+  createServer,
+  caseNames,
+  ERROR_MESSAGES,
+  PORT,
```
pawel-peksa, line 66: I am not sure if you need to export these information.
Export of createServer should be enough.

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+
+    const [originString, queryCase] = request.url.slice(1).split('?');
+
+    const toCase = queryCase?.split('=')[1];
+
+    const errors = checkValidation(originString, toCase);
+
+    if (errors.length) {
+      response.statusCode = 400;
+      response.statusMessage = 'Status: Bad Request';
```
pawel-peksa, line 44:
```suggestion
      response.statusMessage = 'Bad Request';
```

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+  return errors;
+};
+
+const createServer = () => {
+  const server = http.createServer((request, response) => {
+    response.setHeader('Content-type', 'application/json');
+
+    const [originString, queryCase] = request.url.slice(1).split('?');
+
+    const toCase = queryCase?.split('=')[1];
```
pawel-peksa, line 38: Nitpick:
I am pretty sure your solution works as well. However, good practice is to use `URLSeatchParams` :
```
 const searchParams = new URLSearchParams(queryToCase);
 const targetCase = searchParams.get('toCase');
```

pawel-peksa (changes requested):
Good job! The code is correct and works as expected. I've just added couple of minor comments.

pawel-peksa (approved):
Perfectooo 🥇

PR #485, opened 2023-10-18
Develop

Matylda-Lewandowska:
task solution to node_case-transformer-server


szymonpachucki (approved):
LGTM

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+      );
+    }
+
+    if (!toCase) {
+      // eslint-disable-next-line max-len
+      errors.push('"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".');
+    } else {
+      const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+      if (!supportedCases.includes(toCase.toUpperCase())) {
```
pawel-peksa, line 27: Try to think about the conditions here. There should be a solution without nested If statements and a need for use of `else`.

src/createServer.js
```diff
@@ -1,3 +1,56 @@
...
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+const { sendBadRequest } = require('./sendBadRequest');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    const urlParams = new URL(req.url, `http://${req.headers.host}`);
+
+    const textToConvert = urlParams.pathname.substring(1);
+    const toCase = urlParams.searchParams.get('toCase');
```
pawel-peksa, line 10: You could also use the following approach:
```suggestion
        const [textToConvert, queryToCase] = req.url.slice(1).split('?');
    const searchParams = new URLSearchParams(queryToCase);
    const toCase= searchParams.get('toCase');
```

pawel-peksa (changes requested):
Good job! 💪 I've added two minor comments.

choeqq (changes requested):
Waiting for any updates @Matylda-Lewandowska 👀

pawel-peksa (approved):
Well done! Good job 🥇

PR #496, opened 2023-10-20
add task solution

MateuszSlezak:


kuzgan (approved):
Looks good 👍

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+    const newText = convertToCase(textToConvert, toCase);
+    const result = {
+      originalCase: newText.originalCase,
+      targetCase: toCase,
+      originalText: textToConvert,
+      convertedText: newText.convertedText,
+
+    };
+
+    response.writeHeader(400, { 'Content-Type': 'application/json' });
```
choeqq, line 52: since there are no errors here (all of them were handled above) shouldn't it be a 200? 🤔

choeqq (changes requested):

src/createServer.js
MateuszSlezak: true! ;p

choeqq (approved):

PR #514, opened 2023-11-28
first solution

OlhaKoziuk:


src/validatedRequest.js
```diff
@@ -0,0 +1,18 @@
...
+  const errors = [
+    !textToConvert
+    /* eslint-disable max-len */
+    && { message: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' },
+    !toCase
+    && { message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".' },
+    !(!toCase || possibleCases.includes(toCase.toUpperCase()))
+    && { message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.' },
+  ]
+    .filter(Boolean);
```
adioskv, line 13: Never seen that before. Interesting! 😎

src/createServer.js
```diff
@@ -1,3 +1,51 @@
...
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { handleFaviconRequest } = require('./faviconRequest');
+const { validateRequest } = require('./validatedRequest');
+const { createErrorResponse } = require('./errorResponse');
+const { createSuccessResponse } = require('./successResponse');
+const { createJsonResponse } = require('./jsonPesponse');
+
+const PORT = 4000;
```
adioskv, line 9: As i remember, we don't need to hardcode our PORT, we are hardcoding in in `main.js`

adioskv (commented):
### Excellent job! 🚀

adioskv (approved):

src/createServer.js
OlhaKoziuk: Oh, I didnt see this file) just write it to work on my laptop, not exacly understant how it should be in the best way)

StasSokolov (approved):
### Looks really good🔥

PR #516, opened 2023-11-28
Solution

adioskv:


src/getResponse.js
```diff
@@ -0,0 +1,19 @@
+const getResponse = (response, code, body) => {
+  switch (code) {
+    case 400:
```
StasSokolov, line 3: let's status codes to constants and reuse

StasSokolov (approved):
### Looks like masterpiece, i'm approving it but please pay attention at one minor comment🔥

PR #533, opened 2023-11-30
solution1

A1eeex:


Kuper18 (approved):
Nice work) 🔥

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+/* eslint-disable max-len */
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+const { validateError } = require('./validateError');
+
+function createServer() {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-type', 'application/json');
+
+    const parsedUrl = new URL(req.url, 'http://localhost:5700');
```
vasylmiskiv, line 10: Should add a variable for port and save it .env

src/createServer.js
```diff
@@ -1,3 +1,39 @@
...
+        originalText,
+        convertedText,
+      };
+
+      res.statusCode = 200;
+      res.statusMessage = 'Ok';
+      res.end(JSON.stringify(result));
+    }
+
+    if (validateErrorMessage.errors.length > 0) {
```
vasylmiskiv, line 30: Would be better to first check if some errors exist should throw an error, otherwise return the response with result

vasylmiskiv (approved):
Looks fine but pay attention to the comments

PR #534, opened 2023-11-30
add task solution

tilugdmytro:


YaroslavYarynych (approved):
It's a masterpiece🔥🔥🔥

StasSokolov (approved):
### Agree with Yaroslav, looks like masterpiece🔥🔥

PR #538, opened 2023-11-30
add solution case transformer

Kuper18:
Hi, here is my solution!


StasSokolov (approved):
### Great solution, i'm approving it 🔥

PR #562, opened 2024-01-09
added task solution

Yevhen-Kolosov:


src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+
+const BASE = 'http://localhost:5701';
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const reqUrl = new URL(BASE + req.url, BASE);
+    const originalText = reqUrl.pathname.slice(1);
+    const targetCase = reqUrl.searchParams.get('toCase');
```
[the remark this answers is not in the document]
roman-mirzoian: Good point

src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+
+    const errors = [];
+
+    let conversionResult = {
+      originalCase: null,
+      convertedText: null,
+    };
+
+    if (!originalText) {
```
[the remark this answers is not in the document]
roman-mirzoian: Good point

src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+      res.statusCode = 400;
+      res.statusMessage = 'Bad request';
+
+      res.write(JSON.stringify({
+        errors,
+      }));
+    } else {
+      res.statusCode = 200;
+
+      const resultObj = {
```
roman-mirzoian, line 56: Avoid using such abstract names, give a more detailed naming based on what will be stored in this object.

src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const BASE = 'http://localhost:5701';
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const reqUrl = new URL(BASE + req.url, BASE);
```
roman-mirzoian, line 10: I am not sure that such a link will work correctly, have you tried running your program?
https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

roman-mirzoian (changes requested):
Working with node is no different from the programming we learned earlier: break the code into separate functions, files, create constants - all techniques of good code writing must be used constantly.

src/createServer.js
Yevhen-Kolosov: It works (the URL object builds its params correctly), though I will change the first argument to match the examples.

BudnikOleksii (approved):
👍

PR #564, opened 2024-01-10
task solution

VeDono:


src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+    }
+
+    if (!caseType) {
+      errors.push({
+        message:
+          '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    if (!SUPPORTED_CASES.includes(caseType) && caseType) {
```
BudnikOleksii, line 41: Create a separate function and move it to another file

src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+      errors.push({
+        message:
+          'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 44: Use your constant `SUPPORTED_CASES` to provide info about valid cases

src/createServer.js
```diff
@@ -1,3 +1,66 @@
...
+      });
+    }
+
+    try {
+      if (errors.length > 0) {
+        throw errors;
+      }
```
BudnikOleksii, line 51: You can throw an error if any from your validation funciton

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments that should be fixed

BudnikOleksii (approved):
👍

PR #571, opened 2024-01-11
Solution

YuriySobakar:


Anton-Kuchmasov (approved):
Nice job you did!

src/createServer.js
```diff
@@ -1,3 +1,86 @@
 // Write code here
 // Also, you can create additional files in the src folder
 // and import (require) them here
+
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const PORT = 'http://localhost:3000';
```
BudnikOleksii, line 8: Port is only that specified after `:`. In your case 3000

src/createServer.js
```diff
@@ -1,3 +1,86 @@
...
+        });
+      }
+    } else {
+      error.errors.push({
+        message: erorrMessages.noToCaseParam,
+      });
+    }
+
+    if (originalText === '') {
+      error.errors.push({
```
BudnikOleksii, line 53: - Move validation logic to s separate function and file
- Avoid nested `if` statements and `else` too

src/createServer.js
```diff
@@ -1,3 +1,86 @@
...
+      ...convertedTextData,
+    }));
+
+    return res.end();
+  });
+
+  return server;
+}
+
+// createServer().listen(3000);
```
BudnikOleksii, line 82: Remove comments

BudnikOleksii (changes requested):
Good work 👍 
Just a few moments should be fixed

src/ErrorUtil/validateErrors.js
```diff
@@ -0,0 +1,36 @@
...
+
+  if (targetCase) {
+    if (!isSupportedCase) {
+      error.errors.push({
+        message: erorrMessages.invalidCase,
+      });
+    }
+  } else {
+    error.errors.push({
+      message: erorrMessages.noToCaseParam,
```
BudnikOleksii, line 21: Avoid nested `if` and `else` statements

src/ErrorUtil/validateErrors.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+  }
+
+  if (originalText === '') {
```
BudnikOleksii, line 25:
```suggestion
  if (!originalText) {
```
You don't need to compare with an empty string

BudnikOleksii (changes requested):
Approved, but check the comments for the future

src/ErrorUtil/validateErrors.js
YuriySobakar: ok, but how I can do it in this case. I tried switch case but it doesn't work because I must have an array of errors i they are, not only one.

BudnikOleksii: > ok, but how I can do it in this case. I tried switch case but it doesn't work because I must have an array of errors if they are present, not only one.

```javascript
  if (!targetCase) {
    error.errors.push({
      message: erorrMessages.noToCaseParam,
    });
  }
   
  if (!isSupportedCase) {
    error.errors.push({
      message: erorrMessages.invalidCase,
    });
  }
```

BudnikOleksii (approved):

PR #572, opened 2024-01-12
solution

YehorHeneralov:


src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    const newUrl = new URL(req.url, `http://localhost:${PORT}`);
+    const originalText = newUrl.pathname.slice(1);
+    const targetCase = newUrl.searchParams.get('toCase');
+    const errorsArray = [];
+
+    res.setHeader('Content-Type', 'application/json');
+    res.statusCode = 200;
```
MrHunterBoi, line 23: You can't be sure that the request will succeed without errors. It's better to assign success code after all checks have passed

src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+
+    if (!targetCase) {
+      errorsArray.push({
+        message: messages.insertQueryParam,
+      });
+    } else if (!typeOfCases.includes(targetCase)) {
+      errorsArray.push({
+        message: messages.notCorrectCase,
+      });
+    }
```
MrHunterBoi, line 39: You could also move these checks to a separate function

MrHunterBoi (approved):

MrHunterBoi:
Well done

src/createServer.js
```diff
@@ -1,3 +1,72 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase.js');
+
+const typeOfCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+const messages = {
```
BudnikOleksii, line 7: It looks like there aren't just messages, but error messages

src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+  putText: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  insertQueryParam: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  notCorrectCase: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 10: You have a variable `typeOfCases` use it to show valid cases

src/createServer.js
BudnikOleksii: > You can't be sure that the request will succeed without errors. It's better to assign success code after all checks have passed

Good point

src/createServer.js
BudnikOleksii: > You could also move these checks to a separate function

Good point 👍 
- And avoid `else if`

BudnikOleksii (changes requested):
Good work 👍 
Just a few fixes are required

BudnikOleksii (approved):
👍

PR #579, opened 2024-01-15
task done

olegnovikov10:


src/validation.js
```diff
@@ -0,0 +1,36 @@
+const CASES = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+const validate = (text, caseName) => {
+  const errors = [];
+
+  if (text === '') {
```
BudnikOleksii, line 5:
```suggestion
  if (!text) {
```
You don't need to compare with an empty string, just check if exists

src/validation.js
```diff
@@ -0,0 +1,36 @@
...
+        'Text to convert is required.'
+        + ' Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!CASES.includes(caseName) && caseName) {
+    errors.push({
+      message:
+        'This case is not supported. '
+        + 'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 17: You have an array `CASES`, use it to show all available

src/validation.js
```diff
@@ -0,0 +1,36 @@
...
+    });
+  }
+
+  if (caseName === null) {
```
BudnikOleksii, line 21:
```suggestion
  if (!caseName) {
```
The same

BudnikOleksii (changes requested):
Good job 👍 
Left a few suggestions, check them

roman-mirzoian (approved):
Well done 👍

PR #590, opened 2024-01-24
server implemented

jacekkosiba:


MarcinLigmann (approved):
No comments, everything as it should be.

src/utils.js
```diff
@@ -0,0 +1,38 @@
...
+    errors.push({
+      // eslint-disable-next-line max-len
+      message: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  } else if (!cases.includes(targetCase)) {
+    errors.push({
+      // eslint-disable-next-line max-len
+      message: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+    });
+  }
```
mvjl000, line 30: It would be good, to put those error messages into some object and then do:
`message: VALIDATION_ERRORS.NO_TEXT`

src/createServer.js
```diff
@@ -1,3 +1,29 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const { parseRequestUrl, validateData } = require('./utils');
+const { sendErrorResponse, sendSuccessResponse } = require('./reponseHandlers');
+
+function createServer() {
+  const server = http.createServer((request, response) => {
+    if (request.url !== '/favicon.ico') {
```
mvjl000, line 7: How about reverting this condition? The rest of the function wouldn't have to be nested
```suggestion
    if (request.url === '/favicon.ico') return;
```

src/reponseHandlers.js
```diff
@@ -0,0 +1,29 @@
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+function sendErrorResponse(response, errors) {
+  response.statusCode = 400;
+  response.statusMessage = 'Bad request';
```
mvjl000, line 5: I would create some kind of config file, which has all codes and messages, and then use it throughout the project
```javascript
const BAD_REQUEST = {
  code: 400,
  message: 'Bad request'
};
```

src/utils.js
mvjl000: This error messages could be in the same file as response codes mentioned above

mvjl000 (changes requested):
Looks good to me. I have some small improvements though ;)

mpiskadlo (changes requested):
I used wrong account, sorry about that 😆

danon321 (approved):

PR #595, opened 2024-01-25
solution

aureliagorzna:


main.js
```diff
@@ -1,7 +1,3 @@
 const { createServer } = require('./src/createServer');
 
-createServer()
-  .listen(5700, () => {
-    // eslint-disable-next-line no-console
-    console.log('Server started! 🚀');
-  });
+createServer().listen(5700);
```
danon321, line 3: Why u delete this cool information about Server start :////
You can store information about port in variable and first check if it isnt already set in .env file. Like:
const PORT = process.env.PORT || 5700;

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
-// and import (require) them here
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+const { findErrors } = require('./validation');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-type', 'application/json');
+
+    const newURL = new URL(req.url, 'http://localhost:5700');
```
danon321, line 9: You can take this "localhost:5700' information from req.headers. But u have to add 'http://' by yourself.

Its good practise because if u want to run it on some custom domain u dont have change anything in code

src/validation.js
```diff
@@ -0,0 +1,30 @@
...
+const errorMessages = {
+  missingToCase: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  textNotFound: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  incorrectCase: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+};
+
+const findErrors = (text, caseName) => {
+  const errors = [];
+
+  if (!text || text === '') {
```
danon321, line 13: text === '' is useless because !text already handle that

danon321 (changes requested):

mbruhler (approved):

PR #596, opened 2024-01-25
created http server...

kate-prakofyeva:
created http server with working converting function and error handing object

### DIFF
--- a/main.js
+++ b/main.js
@@ -1,7 +1,9 @@
 const { createServer } = require('./src/createServer');
 
+const PORT = process.env.PORT || 5700;
+
 createServer()
   .listen(5700, () => {
     // eslint-disable-next-line no-console
-    console.log('Server started! 🚀');
+    console.log(`Server started! 🚀 port: ${PORT}`);
   });
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -1,3 +1,71 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+/* eslint-disable max-len */
+/* eslint-disable no-console */
+const http = require('http');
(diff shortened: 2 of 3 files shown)


main.js
```diff
@@ -1,7 +1,9 @@
 const { createServer } = require('./src/createServer');
 
+const PORT = process.env.PORT || 5700;
+
 createServer()
   .listen(5700, () => {
```
danon321, line 6: If you have that PORT const why u didnt use it there?

src/createServer.js
```diff
@@ -1,3 +1,71 @@
...
+/* eslint-disable no-console */
+const http = require('http');
+const { convertToCase } = require('./convertToCase/convertToCase');
+
+const createServer = () => {
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const errorHanding = {
+      errors: [
```
danon321, line 11: It's necessary to do this like that? I mean errors array inside a object. Just const error = []; will be not enough?

danon321 (changes requested):

src/createServer.js
kate-prakofyeva: From Readme file 
```
#### Validation
Text in the URL and query param `toCase` are mandatory. Also, `toCase` value should be one of the supported cases.
If something is not correct, you should respond with 400 status, `Bad request` statusText and the next payload:
```json
{
  "errors": [
    {
      "message": "<SPECIFIC MESSAGE TEXT HERE>"
    }
  ],
}
```
Array of messages can contain more than one error. For example, if both text and case are not provided.
```

This why I decided to do it in this way

danon321 (approved):

PR #602, opened 2024-02-03
Solution

Hladkyi03:


src/requestHandlers/errorHandler.js
```diff
@@ -0,0 +1,30 @@
...
+    errors.push({
+      message: 'Text to convert is required. Correct request is: '
+        + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!availableCases.includes(targetCase) && targetCase) {
+    errors.push({
+      message: 'This case is not supported. Available cases: '
+        + 'SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
vadiimvooo, line 16:
```suggestion
        + availableCases.join(", "),
```

src/convertToCase/toWords.js
```diff
@@ -7,6 +7,10 @@
  * @returns {string[]}
  */
 function toWords(text, originalCase) {
+  if (!text) {
+    throw new Error('Text should not be null');
+  }
+
```
vadiimvooo, line 13: It's not necessary to modify this method.

src/requestHandlers/errorHandler.js
vadiimvooo: Or you can also use some object to store by some keys Error text and just show here it.
```
const VALIDATION_ERRORS = {
  NO_TEXT: 'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
  NO_TARGET_CASE: '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase= 
 <CASE_NAME>".',
  CASE_NOT_SUPPORTED: 'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
};
```

src/createServer.js
```diff
@@ -1,3 +1,38 @@
...
+
+function createServer() {
+  const server = http.createServer(async(req, res) => {
+    res.setHeader('Content-type', 'application/json');
+
+    const normalizedURL = new URL(req.url, 'http://localhost:5700');
+
+    const originalText = normalizedURL.pathname.slice(1);
+    const targetCase = normalizedURL.searchParams.get('toCase');
+
```
vadiimvooo, line 16: Here you can make validation if data which comes from request is valid and if not you can return an error. Else you can use convertedToCase func.

vadiimvooo (changes requested):

SanyaBratashchuk (approved):

PR #604, opened 2024-02-05
added case-transformer-server

PonomarchukLana:


src/createServer.js
```diff
@@ -1,3 +1,62 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+'use strict';
+
+const http = require('http');
+const { convertToCase } = require('./convertToCase');
+
+function createServer() {
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
roman-mirzoian, line 7: Static constants/data are best kept outside the function definition.

src/createServer.js
```diff
@@ -1,3 +1,62 @@
...
+    const textToConvert = url[0].slice(1);
+    const params = new URLSearchParams(url[1]);
+    const toCase = params.get('toCase');
+    const errors = [];
+
+    if (!textToConvert) {
+      errors.push({
+        message:
+          'Text to convert is required. '
+          + 'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
roman-mirzoian, line 21: It is more convenient to keep the error text in separate constants or an object with the corresponding keys.

roman-mirzoian (approved):
Well done, but pay attention to the comments 👍

PR #627, opened 2024-03-08
need help

denys-khvashchenko:
do not understand why error checking doesn't work correctly

### DIFF
--- a/src/createServer.js
+++ b/src/createServer.js
@@ -14,17 +14,17 @@ const createServer = () => {
     const targetCase = new URLSearchParams(inputCase).get('toCase');
     const errors = messageError(originalText, targetCase);
 
-    const {
-      originalCase,
-      convertedText,
-    } = convertToCase(originalText, targetCase);
-
     if (errors.length) {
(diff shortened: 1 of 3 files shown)


src/createServer.js
```diff
@@ -1,3 +1,46 @@
...
+
+    const [inputText, inputCase] = request.url.split('?');
+    const originalText = inputText.slice(1);
+    const targetCase = new URLSearchParams(inputCase).get('toCase');
+    const errors = messageError(originalText, targetCase);
+
+    const {
+      originalCase,
+      convertedText,
+    } = convertToCase(originalText, targetCase);
```
lerastarynets, line 20: you see errors, because u don't validate `targetCase`, so its nullish/invalid value goes to this function, so an error should be thrown before the execution of `convertToCase`

consider creating some validations for targetCase value before this row

lerastarynets (changes requested):
Nice start! Left comments that should help you make code work)

anastasiiavorobiova (approved):
Great work!

PR #638, opened 2024-03-12
add task solution

MaxKalalashnyk02:


src/createServer.js
```diff
@@ -1,3 +1,50 @@
...
+    const text = req.url.split('?')[0].slice(1);
+
+    try {
+      const errors = getErrors(text, toCase);
+
+      if (errors.length > 0) {
+        throw errors;
+      }
+
+      res.writeHead(200, 'OK', { 'Content-Type': 'application/json' });
```
DanilWeda, line 24: Can describe all status codes in constants ( `200,400,500`)

DanilWeda (changes requested):
Good job! 
Check small comments for improvement. 
Thanks)

PR #643, opened 2024-03-20
impl solution

DimasSavchenko:


src/validateCaseAndText.js
```diff
@@ -0,0 +1,28 @@
+const validateCaseAndText = (caseType, text) => {
+  const errorMessages = [];
+  const validCaseTypes = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+  const validCaseTypesString = validCaseTypes.toString().replaceAll(',', ', ');
```
vadiimvooo, line 4:
```suggestion
  const validCaseTypesString = validCaseTypes.join(', ');
```

vadiimvooo (changes requested):
Good job. You need to fix your code to pass tests. Please make corrections provided below.

src/validateCaseAndText.js
DimasSavchenko: Sure, thx :)

l4st1m0za (approved):
Good job!

PR #648, opened 2024-04-05
add solution

HolyRaveOS:


src/createServer.js
```diff
@@ -1,3 +1,60 @@
-// Write code here
-// Also, you can create additional files in the src folder
-// and import (require) them here
+const http = require('http');
+const url = require('url');
+const { convertToCase } = require('./convertToCase');
+const casesToConvert = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
BudnikOleksii, line 4: Constants should be in upper case

src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+
+    if (!toCase) {
+      errors.push({
+        message: `"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".`,
+      });
+    } else if (!casesToConvert.includes(toCase.toUpperCase())) {
+      errors.push({
+        message: `This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.`,
+      });
+    }
```
BudnikOleksii, line 30: - Create a function to get errors and move it to a separate file.
- Create an object with error messages
- "Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER." you can see all available cases in the `casesToConvert` variable so use it in the message as well

BudnikOleksii (changes requested):
Good work 👍 
Let's create a separate function for errors

StasSokolov1 (approved):
Great job🔥

PR #650, opened 2024-04-11
add task solution

PodvAx:


src/validate.js
```diff
@@ -0,0 +1,50 @@
...
+  'Text to convert is required. ' +
+  'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+
+const noParamsErrorMessage =
+  '"toCase" query param is required. ' +
+  'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+
+const invalidCaseErrorMessage =
+  'This case is not supported. ' +
+  'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.';
```
DarkMistyRoom, line 13: would be great to combine the error messages into an object and move all constants to a separate file

src/validate.js
```diff
@@ -0,0 +1,50 @@
...
+  return { message: message };
+}
+
+function validate(textToConvert, toCase) {
+  const response = {};
+  const errors = [];
+
+  const isInvalidCase = !validCases.includes(toCase);
+
+  if (!textToConvert || !toCase || isInvalidCase) {
```
DarkMistyRoom, line 25: no need to use this `if` as you check the same conditions in the nested `if`s

DarkMistyRoom (changes requested):
Good job! But let's improve it a bit

PR #663, opened 2024-05-02
Node case transformer server

nk8bnj:


src/createServer.js
```diff
@@ -1,3 +1,45 @@
...
+
+      const splittedUrl = req.url.split('?');
+      const originalText = splittedUrl[0].slice(1);
+      const targetCase = new URLSearchParams(splittedUrl[1]).get('toCase');
+
+      errors = getErrors(originalText, targetCase);
+
+      if (errors.length) {
+        throw new Error();
+      }
```
BudnikOleksii, line 20: You can throw errors in `getErrors` function if any so you don't need to do it here manually

src/utils/getErrors.js
```diff
@@ -0,0 +1,28 @@
...
+      message: `"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".`,
+    });
+  }
+
+  if (
+    toCase &&
+    !['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'].includes(toCase)
+  ) {
+    errors.push({
+      message: `This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.`,
```
BudnikOleksii, line 21: Create a constant for `['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER']` and use it to generate message with available cases

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

BudnikOleksii (approved):
👍

PR #668, opened 2024-05-07
Add solution

ViktoriiaRepo:


src/validator.js
```diff
@@ -0,0 +1,36 @@
+/* eslint-disable max-len */
+const validator = (text, caseToTransform) => {
+  const errors = [];
+  const availableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
BudnikOleksii, line 4: Move `availableCases` outside the function and the constant should be in the upper case

src/validator.js
```diff
@@ -0,0 +1,36 @@
...
+    errors.push({
+      message:
+        '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!availableCases.includes(caseToTransform) && caseToTransform) {
+    errors.push({
+      message:
+        'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 23: Use `supportedCases` to generate a message with available cases

BudnikOleksii (changes requested):
Good work 👍
Just a few fixes are required

BudnikOleksii (approved):
👍

PR #669, opened 2024-05-07
Solution

winniepuhh:


src/errorMessage.js
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable max-len */
+const errorMessage = (text, toCase) => {
+  const errors = [];
+  const availableCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
BudnikOleksii, line 4: - Function name should start from verb
- Move `availableCases` outside the function and the constant should be in the upper case

src/errorMessage.js
```diff
@@ -0,0 +1,34 @@
...
+    errors.push({
+      message:
+        '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!availableCases.includes(toCase) && toCase) {
+    errors.push({
+      message:
+        'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 23: Use `supportedCases` to generate a message with available cases

BudnikOleksii (changes requested):
Good work 👍
Just a few fixes are required

BudnikOleksii (approved):
👍

PR #671, opened 2024-05-08
add task solution

vladarskyi:


src/validateParams.js
```diff
@@ -0,0 +1,35 @@
...
+        // eslint-disable-next-line max-len
+        '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (!CASES.includes(toCase) && toCase) {
+    errors.push({
+      message:
+        // eslint-disable-next-line max-len
+        'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 26: Use `CASES` variable to generate an error message

BudnikOleksii (approved):
👍

src/validateParams.js
vladarskyi: I'm sorry, I didn't catch it. Do you mean create new variable to save the messages? Couse I don't understand how can I use `CASES` to this.

PR #672, opened 2024-05-08
solution

ivan-gritsenko:


src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+
+    if (!toCase) {
+      errors.push({
+        message: errorTypes.toCaseRequired,
+      });
+    } else if (!CASES.has(toCase.toUpperCase())) {
+      errors.push({
+        message: errorTypes.toCaseNotSupported,
+      });
+    }
```
BudnikOleksii, line 23: Create a function that validates input for errors and move it to a separate file

src/createServer.js
```diff
@@ -1,3 +1,67 @@
...
+    );
+
+    const response = {
+      originalCase,
+      targetCase: toCase,
+      originalText: textToConvert,
+      convertedText,
+    };
+
+    if (originalCase !== null && convertedText !== null) {
```
BudnikOleksii, line 52: You don't need to compare with `null` just check if a value exists
https://developer.mozilla.org/en-US/docs/Glossary/Truthy

src/errorTypes.js
```diff
@@ -0,0 +1,12 @@
+/* eslint-disable max-len */
+const errorTypes = {
+  toCaseRequired:
+    '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+  toCaseNotSupported:
+    'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
BudnikOleksii, line 6: Generate the message with "Available cases" depending on `CASES` variable

BudnikOleksii (changes requested):
Good work 👍
Just a few fixes are required

BudnikOleksii (approved):
👍

PR #674, opened 2024-05-08
add solution

illya-onyshchuk:


src/validateCase.js
```diff
@@ -0,0 +1,7 @@
+function validateCase(caseName) {
+  return ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'].includes(caseName);
```
BudnikOleksii, line 2: Create a constant for `['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER']` and use it to generate an error message with available cases

BudnikOleksii (approved):
👍

PR #677, opened 2024-05-08
add solution

Vlobdam:


src/createServer.js
```diff
@@ -1,3 +1,74 @@
...
+
+const MISSING_TEXT =
+  // eslint-disable-next-line max-len
+  'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+const MISSING_CASE =
+  // eslint-disable-next-line max-len
+  '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".';
+const UNSUPPORTED_CASE =
+  // eslint-disable-next-line max-len
+  'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.';
```
BudnikOleksii, line 13: Generate the message with "Available cases" depending on `SUPPORTED_CASES ` variable

src/createServer.js
```diff
@@ -1,3 +1,74 @@
...
+
+      if (!toCase) {
+        body.errors.push({
+          message: MISSING_CASE,
+        });
+      } else if (!SUPPORTED_CASES.includes(toCase)) {
+        body.errors.push({
+          message: UNSUPPORTED_CASE,
+        });
+      }
```
BudnikOleksii, line 52: - Create a function that validates input for errors and move it to a separate file
- statusCode = 200 should be only in case without errors

BudnikOleksii (changes requested):
Good work 👍
Just a few fixes are required

BudnikOleksii (approved):
👍

PR #681, opened 2024-05-09
Solution

rafmamedov:


src/utils/helpers.js
```diff
@@ -0,0 +1,66 @@
...
+
+  if (
+    targetCase &&
+    !supportedCases.some(
+      (supportedCase) => supportedCase === targetCase.toUpperCase(),
+    )
+  ) {
+    errors.push({
+      message: `This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.`,
+    });
```
DarkMistyRoom, line 55: Also would be great to create an object with all error messages as constant

DarkMistyRoom (changes requested):
Good job! Only one small suggestion

src/utils/helpers.js
rafmamedov: Fixed it, thanks!

lerastarynets (approved):
lgtm 🚀

PR #683, opened 2024-05-09
task solution

Oleksandr2108:


src/createServer.js
```diff
@@ -1,3 +1,60 @@
...
+    if (
+      targetCase &&
+      !['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'].includes(
+        targetCase.toUpperCase(),
+      )
+    ) {
+      errors.push({
+        message: `This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.`,
+      });
+    }
```
BudnikOleksii, line 33: - Create a function that will validate data and move it to a separate file
- Create a constant for available cases - `['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER']`, and also use it to generate an error message regarding available cases

BudnikOleksii (changes requested):
Good work 👍
Just a few fixes are required

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+  const server = http.createServer((req, res) => {
+    res.setHeader('Content-Type', 'application/json');
+
+    const [originalText, splitedURL] = req.url.slice(1).split('?');
+    const targetCase = new URLSearchParams(splitedURL).get('toCase');
+
+    const errors = validateRequest(originalText, targetCase);
+
+    if (errors.length > 0) {
+      res.statusCode = 400;
```
StasSokolov1, line 15: will be better create a constant with a status codes

src/createServer.js
```diff
@@ -1,3 +1,40 @@
...
+    );
+
+    const result = {
+      originalCase,
+      targetCase,
+      originalText,
+      convertedText,
+    };
+
+    res.statusCode = 200;
```
StasSokolov1, line 33: also here

StasSokolov1 (changes requested):
Almost done, let's fix a few minor comments!

BudnikOleksii (approved):
👍

PR #688, opened 2024-05-14
Solution

oshapkun:


src/validateInput.js
```diff
@@ -0,0 +1,36 @@
+function validateInput(text, caseName) {
+  const errors = [];
+
+  if (!text) {
+    errors.push({
+      message:
+        'Text to convert is required. Correct request is: ' +
+        '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
vadiimvooo, line 8: Use object to save string variables.

vadiimvooo (approved):
Well done. Just one suggestion.

PR #698, opened 2024-06-10
solution

YarynaPuhach:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+
+    if (!toCase) {
+      errors.push({
+        message: '"toCase" query param is required. Correct request is: '
+          + '"/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!validCases.includes(toCase)) {
+      errors.push({
+        message: 'This case is not supported. Available cases: '
+          + 'SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
nataliaklonowska, line 33: for better maintainability instead of using the hardcoded string "SNAKE, KEBAB, CAMEL, PASCAL, UPPER." you should use validCases array. This ensures that the list of valid cases is consistent and easier to update

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #704, opened 2024-06-11
task solution

Kacper7001:


src/validator.js
```diff
@@ -0,0 +1,27 @@
+const { invalidText, invalidToCase, invalidToCaseValue } = require('./error');
+
+const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
nataliaklonowska, line 3: don't define supportedCases a second time because it leads to code duplication making maintenance more difficult and updates more error-prone

nataliaklonowska (changes requested):

choeqq (approved):

PR #707, opened 2024-06-13
Solution

pawelkloc:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+      });
+    }
+
+    const validCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    if (toCase && !validCases.includes(toCase.toUpperCase())) {
+      errors.push({
+        message:
+          'This case is not supported. Available cases: ' +
+          'SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
nataliaklonowska, line 35: for better maintainability instead of using the hardcoded string "SNAKE, KEBAB, CAMEL, PASCAL, UPPER." you should use validCases array. This ensures that the list of valid cases is consistent and easier to update

nataliaklonowska (changes requested):

nataliaklonowska (approved):

PR #712, opened 2024-06-26
Solution

MatheusStefan:


src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+    const errors = [];
+
+    if (!text) {
+      errors.push({
+        message:
+          'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    }
+
+    const supportedCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
pedro-ruas, line 23: This const could be hosted outside of the server function

src/createServer.js
```diff
@@ -1,3 +1,61 @@
...
+          'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
+    }
+
+    if (errors.length) {
+      res.statusCode = 400;
+      res.statusMessage = 'Bad Request';
+      res.end(JSON.stringify({ errors }));
+
+      return;
```
pedro-ruas, line 42: You could return the previous line instead of returning nothing, the code will do the same

pedro-ruas (changes requested):
Just a really small change, great job!

EdPirro (approved):
Great job, @MatheusStefan! Keep it up 🚀

PR #713, opened 2024-06-26
solution

Allanpt:


src/createServer.js
pedro-ruas, line 1: Very nice destructuring approaches! Great job!

src/errors.js
```diff
@@ -0,0 +1,34 @@
+function checkError(originalText, targetCase) {
+  const validCases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
pedro-ruas, line 2: This const could be declared outside the function, as it won't change anyway

src/errors.js
```diff
@@ -0,0 +1,34 @@
...
+
+  if (!targetCase) {
+    errors.push({
+      message:
+        '"toCase" query param is required.' +
+        ' Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (targetCase && !validCases.some((cas) => cas === targetCase)) {
```
pedro-ruas, line 21: This _if_ could be an _else_ of the previous _if_
```
if (!targetCase) {
    ...
} else if (!validCases ...) {
    ...
}
```

And another possible approach would be to change _some_ for _includes_, having the same result

pedro-ruas (changes requested):
Minimal changes required, great job!

jv-aquino (approved):
nice job allan!

PR #715, opened 2024-06-27
adding task solution

eduvinagre:
I've created a file named utils.js so that I could handle the helpers functions (handling error messages and request validations) to make the code more modular and maintainable.


jv-aquino (approved):
Great job Edu, nice use of modular code (utils.js)!!!

PR #719, opened 2024-07-01
add task solution

JohnyJDX:


alexander-ignatow (approved):
Good solution!

Just one thing - it is not critical for the given task, but in the future please consider writing more "modular" code

It makes sense to have separate modules (or at least functions) for input validation/preparation, operation handling, error handling, etc

PR #726, opened 2024-07-03
Implement http server

olafchuszno:


src/createServer.js
```diff
@@ -1,3 +1,100 @@
...
+      request.url,
+      `http://${request.headers.host}`,
+    );
+
+    const searchParams = normalizedURL.searchParams;
+
+    const textToConvert = normalizedURL.pathname.slice(1);
+    const targetCase = searchParams.get('toCase');
+
+    // If no 'text to convert', set error in errors array
```
BogdanMaliuta, line 28: remove commetns from code

BogdanMaliuta (commented):
Not bad solution, but for me here you need to restructurize your if/else for avoiding nesting, and better code readability, for example you can ititialize your responseData.errors before all checks and remove this ` 
if (!Array.isArray(responseData.errors)) {
        responseData.errors = [];
      }`
 from all your if blocks, etc.

BogdanMaliuta (changes requested):
Requesting changes

src/createServer.js
```diff
@@ -1,3 +1,83 @@
...
+    // If no toCase query param
+    if (!searchParams.has('toCase')) {
+      responseData.errors.push({
+        message: `"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".`,
+      });
+
+      // If Case not supported -> Append error
+    } else if (targetCase === null || !supportedCases.includes(targetCase)) {
+      responseData.errors.push({
+        message: `This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.`,
```
natalia-klonowska, line 43: for better maintainability instead of hardcoded "SNAKE, KEBAB, CAMEL, PASCAL, UPPER." in string you should use supportedCases array. This ensures that the list of valid cases is consistent and easier to update

natalia-klonowska (changes requested):
remember to delete comments from code :)

BogdanMaliuta (approved):
Solution looks better!

PR #729, opened 2024-07-03
Task case transformer server GH

HigorPedroso:


src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+      req.url,
+      `http://${req.headers.host}`,
+    );
+    const textToConvert = pathname.slice(1);
+    const targetCase = searchParams.get('toCase');
+
+    const hasAnyErrors = handleErrors(res, textToConvert, targetCase);
+
+    if (hasAnyErrors) {
+      return;
```
jv-aquino, line 24: i think it would be a better practice to use sendError here (also improving code readability

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+    const result = convertToCase(textToConvert, targetCase);
+
+    const response = {
+      originalCase: result.originalCase,
+      targetCase,
+      originalText: textToConvert,
+      convertedText: result.convertedText,
+    };
+
+    res.end(JSON.stringify(response));
```
jv-aquino, line 36: also a better practice would be to manually use 
res.statusCode = 200;
here to clarify the OK response

jv-aquino (changes requested):
There are two really minor practices that i'd like you to remember here - the entire code is fine, just 2 really minor things to remember!

jv-aquino (approved):
Great job Higor!

PR #732, opened 2024-07-04
Develop

yaroslav-na:


src/createServer.js
```diff
@@ -1,3 +1,50 @@
...
+    response.setHeader('Content-Type', 'application/json');
+
+    const errorCollector = getErrorCollector();
+    const url = new URL(request.url, `http://${request.headers.host}`);
+    const originalText = url.pathname.slice(1);
+    const targetCase = url.searchParams.get(TO_CASE_QUERY);
+
+    if (!originalText) {
+      errorCollector.catchIfThrown(() => {
+        throw new Error(ERROR_MESSAGE.TEXT_REQUIRED);
```
anastasiiavorobiova, line 18: No need to throw new errors. Consider using a validation function as a middleware to check the response before the main logic. Or use a try-catch block here

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Great work!

PR #736, opened 2024-07-05
Solution

MarchAlbion:


src/validate.js
```diff
@@ -0,0 +1,30 @@
...
+    errorsArray.push({
+      message:
+        '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    });
+  }
+
+  if (toCase && !cases.includes(toCase)) {
+    errorsArray.push({
+      message:
+        'This case is not supported. Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
```
DanilWeda, line 23: You can move all messages to constants.
Also please fix eslint error.

src/validate.js
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable max-lines */
+function validate(targetWord, toCase) {
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
DanilWeda, line 3: Better way use SUPPORTED_CASES ( or something like that) for constant name.
Also, move all constants (status code, error messages) to upper scope or better way - to separate file.

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+
+    const targetWord = params
+      .split('')
+      .splice(1, params.length - 1)
+      .join('');
+
+    const errors = validate(targetWord, toCase);
+
+    if (errors.length > 0) {
+      res.writeHead(400, { 'Content-Type': 'application/json' });
```
DanilWeda, line 22: You can move all status code to constants

src/createServer.js
```diff
@@ -1,3 +1,42 @@
...
+      res.end(JSON.stringify({ errors }));
+
+      return;
+    }
+
+    const { originalCase, convertedText } = convertToCase(targetWord, toCase);
+
+    res.writeHead(200, { 'Content-Type': 'application/json' });
```
DanilWeda, line 30: You can move all status code to constants

DanilWeda (changes requested):
Good job! 
I left some minor comments. 
Many thanks!)

src/validate.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable max-lines */
+function validate(targetWord, toCase) {
+  const SUPPORTED_CASES = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+  const errorsArray = [];
+  const TEXT_TO_CONVERT_ERROR =
```
DanilWeda, line 5: All this constants need to be in one place. Move it to separate file, then import. 
Or move outside function without status code ( if you not create separate file for all constants) 
const STATUS_CODE_400 
const STATUS_CODE_200
const TEXT_TO_CONVERT_ERROR
const SUPPORTED_CASES

DanilWeda (changes requested):
Good job! 
Check minor comment about constants. 
Many thanks )

DanilWeda (approved):
Good job. 
Approved!
Many thanks)

PR #741, opened 2024-07-08
solution

lmmaarriiaa:


src/validation.js
```diff
@@ -0,0 +1,38 @@
...
+      message:
+        'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+    };
+
+    errors.push(message);
+  }
+
+  if (!toCase) {
+    const message = {
+      message:
```
DanilWeda, line 17: You can move all messages to constants

src/createServer.js
```diff
@@ -1,3 +1,44 @@
...
+    const toCase = params.get('toCase') || '';
+    const textToConvert = text.slice(1);
+
+    const errors = validateParams(textToConvert, toCase);
+
+    if (errors.length > 0) {
+      const data = JSON.stringify({ errors });
+
+      res.setHeader('Content-Type', 'application/json');
+      res.statusCode = 400;
```
DanilWeda, line 20: You can move all status code to constants

src/validation.js
```diff
@@ -0,0 +1,38 @@
+/* eslint-disable max-len */
+const validateParams = (textToConvert, toCase) => {
+  const cases = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
```
DanilWeda, line 3: Better way use CASES_TO_CONVERT ( or something like that) for constant name.
Also, move all constants (status code, error messages) to the upper scope or better way - to separate file.

DanilWeda (changes requested):
Good job. 
Please check some comments that I left. 
Many thanks)

src/validation.js
lmmaarriiaa: Hi! 
Fixed everything, thanks!

DanilWeda (approved):
Great job!
Approved. 
Thanks)

PR #759, opened 2024-08-16
Develop

Patryk-Buczkowski:


src/createServer.js
```diff
@@ -1,3 +1,64 @@
...
+        message:
+          '"toCase" query param is required. ' +
+          'Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (!cases.includes(toCase)) {
+      errors.push({
+        message:
+          'This case is not supported. ' +
+          'Available cases: SNAKE, KEBAB, CAMEL, PASCAL, UPPER.',
+      });
```
natalia-klonowska, line 35: for better maintainability instead of using the hardcoded string "SNAKE, KEBAB, CAMEL, PASCAL, UPPER." you should use cases array. This ensures that the list of valid cases is consistent and easier to update

natalia-klonowska (changes requested):
Great work 😃

natalia-klonowska (approved):

PR #770, opened 2024-09-21
 node_case-transformer-server

NataliaTopornitska:


src/createServer.js
```diff
@@ -1,3 +1,68 @@
...
+
+    const userText = url.pathname.slice(1);
+    const typeOfCase = url.searchParams.get('toCase');
+    const errors = [];
+    const typesOfCase = ['SNAKE', 'KEBAB', 'CAMEL', 'PASCAL', 'UPPER'];
+
+    if (!userText) {
+      errors.push({
+        message:
+          'Text to convert is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
```
vadiimvooo, line 22: Create some object which will store messages with specific keys and reuse it here.

vadiimvooo (approved):
Well done, just one small suggestion.

PR #823, opened 2025-01-10
add solution

causeurgnocchi:


src/validateUrl.js
```diff
@@ -0,0 +1,35 @@
+function validateURL(url) {
```
joaorpereira: this file is not been used.

joaorpereira (changes requested):
GREAT work all good. 

Just remove the `validateURL` file or `validateURL` code from `createServer.js` file and use the imported file.

src/validateUrl.js
causeurgnocchi: João, thank you for taking your time reviewing my code.

The code has been adjusted.

joaorpereira (approved):

PR #894, opened 2025-04-15
Solution for case-transformer-server

kovaltar:


kovaltar:
@mateacademy-ai-mentor the originalCase is in conversion object

anastasiiavorobiova (approved):
Great work!

PR #909, opened 2025-05-20
feat: add task solution

s1rserg:


s1rserg:
@mateacademy-ai-mentor 
In fact, URLSearchParams(undefined) does not throw an error. It simply creates an empty URLSearchParams object. recheck the documentation

PR #957, opened 2025-08-19
add solution

zbvika:


zbvika:
@mateacademy-ai-mentor Thank you for the review. However, I ask you to check again, since in the original code the convertToCase function takes text first, and then caseName.

zbvika:
@mateacademy-ai-mentor I tried changing the order, but the tests don’t pass because they are set up so that convertToCase takes text first, and then caseName.

PR #969, opened 2025-09-01
case-transformer-server solution

NadiiaBulmak:


src/createServer.js
```diff
@@ -1,3 +1,72 @@
...
+        message:
+          // eslint-disable-next-line max-len
+          '"toCase" query param is required. Correct request is: "/<TEXT_TO_CONVERT>?toCase=<CASE_NAME>".',
+      });
+    } else if (
+      targetCase !== 'SNAKE' &&
+      targetCase !== 'KEBAB' &&
+      targetCase !== 'CAMEL' &&
+      targetCase !== 'PASCAL' &&
+      targetCase !== 'UPPER'
```
Anton-Kuchmasov: Consider creating an enum/type (even you can use `object as const` if needed) and use it to prevent hard-coded errors

Anton-Kuchmasov (approved):
Good job!
Please check comment below

src/createServer.js
NadiiaBulmak: Thank you for your recommendations! It is valuable for me! 💫

PR #970, opened 2025-09-03
finish task

nbryk:


nbryk:
@mateacademy-ai-mentor There is no error in the code - it is an incorrect description in the task. The code works correctly.

PR #1044, opened 2026-01-13
Dev

DrDakka:


DrDakka:
1. Fixed issue with unsupported case logics 
2. a single lower-case word can be any: kebab-case, snake-case, camel-case :)
'test' word fits well for any of them. Main thing - the code works well, and fit's requirements of README.MD

DrDakka:
> #A better approach would be to check for delimiters first. If the string includes('_'), validate it as SNAKE. If it includes('-'), validate it as KEBAB. If it has neither and is all lowercase, you can then classify it as camelCase."

Without separators in the string, you can't tell if a word is kebab, camel or snake case. Same logics about '-' and '_' can be used for camelCase: we need at least 1 capital letter to determine it. So i don't see error.

PR #1084, opened 2026-03-21
feat: implementa case-transformer server e documentação completa

glauccoeng-prog:


raulriato (approved):
Your code looks fine. Just a tip for next projects:

It's not good practice to leave comments on the final code.

